PR #13, opened 2016-04-08
Enable CloudFoundry deployment natively

hochm:
Allows to deploy to a CF environment natively, be getting the port dynamically out of the env.


andris9:
There is no need to add support for different environment variables as these can be linked from the config files, see [here](https://github.com/andris9/mailtrain/tree/v1.1#using-environment-variables)

PR #15, opened 2016-04-12
Add body size request limit in config files

cauliturtle:
during days for testing, while sending large batch for mails via SMTP hosts and webhooks enabled. post request of webhook will be 413 entity too large. 

so here is the fix


andris9:
Good idea. I added this feature with a47e87d7bc2a62fb4d71c00f03ba673533ed6829

PR #57, opened 2016-06-02
Added node for MAMP users

wesbos:
Struggled with this for about 45 minutes and it seems its something that trips up most MAMP users when they run mysql outside of the mamp PHP environment. 

Not sure if this is too much info for the config file, so feel free to not accept this :)


andris9:
Thanks! I installed MAMP myself as well because someone else also complained about it but never got into actually testing it to find out what is going on.

PR #70, opened 2016-06-22
Delete subscription endpoint for API

macarrillo:
We need it to synchronise all of our users via the API and found some cases where we would like to update the email of a subscription or delete some of them. This endpoint will allow us to do so by deleting old subscriptions and, if needed, replace them with new ones. I replicated the same pattern used for the unsubscribe feature into this new api endpoint, which required a small modification to the model.

Please feel free to comment. It would be great to have this feature supported in the original repository. I'll address any comments that you have. :)

cc @MaheshCasiraghi

### DIFF
--- a/lib/models/subscriptions.js
+++ b/lib/models/subscriptions.js
@@ -825,16 +825,16 @@ module.exports.changeStatus = (id, listId, campaignId, status, callback) => {
     });
 };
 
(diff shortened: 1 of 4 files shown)


andris9:
I did not want to update the .delete call to take email address as an argument, so I modified the code from this PR. Deleting a subscriber is available from v1.13.0

PR #85, opened 2016-08-11
Add LDAP authentication backend

nebulade:
Hi, this pull request is not meant to get merged at the moment. I just use this to see if this is useful to get upstream in the first place. 

The LDAP auth backend is meant for the https://cloudron.io app, on which I am working at.

So far it is functional during my rudimentary testing. 

Currently this still depends on user records in the users db table, to have a place to store the API accessToken. LDAP user records are inserted with an empty password. Not sure if that is the way to go here. Also the admin record is inserted by `mailtrain.sql` which might not be the best idea when only using LDAP auth.

Additionally I am working on disabling the UI bits for password reset and change.

Would be great to get some feedback and also get to know if this even has any chance to get upstreamed.

Thanks for the great application already! :-)

### DIFF
--- a/app.js
+++ b/app.js
@@ -135,6 +135,10 @@ passport.setup(app);
 app.use((req, res, next) => {
     res.locals.flash = req.flash.bind(req);
     res.locals.user = req.user;
+    res.locals.ldap = {
+        enabled: config.ldap.enabled,
+        passwordresetlink: config.ldap.passwordresetlink
+    };
 
     let menu = [{
         title: 'Home',
--- a/config/default.toml
+++ b/config/default.toml
@@ -87,3 +87,11 @@ host="0.0.0.0"
 username="testuser"
 password="testpass"
 logger=false
+
+[ldap]
+# enable to use ldap user backend
+enabled=false
+host="localhost"
+port=3002
+baseDN="ou=users,dc=company"
+filter="(|(username={{username}})(mail={{username}}))"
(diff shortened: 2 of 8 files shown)


andris9:
Awesome idea!

The password mismatch between LDAP and local should not be an issue as the password from the database is not checked when performing LDAP auth, so the PR seems quite OK as it already is.

Maybe also add a check to the local auth so that an empty password could not be used. This would prevent the situation where you disable LDAP for whatever reason and now everybody can sign in using a blank password.

nebulade:
Good idea, will update the pull request in a bit. Just working through the html templates to hide the non relevant parts, when using LDAP

nebulade:
I have now added an additional config variable to specify the password reset link, so users can proceed to the LDAP management UI

andris9:
Great! I'll set up a local LDAP server later today and try it out

gramakri:
@andris9 did you get a chance to try it out?

nebulade:
Is there anything I can help with getting test done for this feature?

andris9:
Actually I did not manage to get it working. Could you provide a sample server script (probably using ldapjs) as well that can be used as test auth backend?

Something like this:

```
node test-ldap-server.js &
node index.js
```

nebulade:
Yeah LDAP can  be quite tricky to start with. I quickly built an ldapjs based one and it's now at https://github.com/nebulade/ldapjstestserver
```
npm install
./index.js
```
It just has two hardcoded user records, which it prints out, very basic code.
Should hopefully work with the configuration bits in mailtrain would look like:
```
[ldap]
# enable to use ldap user backend
enabled=true
host="localhost"
port=3002
baseDN="ou=users,dc=example"
filter="(|(username={{username}})(mail={{username}}))"
passwordresetlink="http://example.com"
```

andris9:
Thanks, I'll check it out in a day or two

PR #88, opened 2016-08-27
Dockerization attempt

djalal:
This is a third install way for Mailtrain, i.e based on Docker containers only.

This opens also automated docker images builds, that is for distribution through Docker Hub (https://hub.docker.com)

I added the `docker-compose.yml` with minimum setup, except the nifty MailCatcher (https://mailcatcher.me/) component to vizualize mails as they are sent.

This proved very useful when evaluating MailTrain, since NodeJS is not my main playground and I did not feel the need to install all the pre-requisites.

Hope this helps!


andris9:
Sorry for the long delay, I wasn't sure how to respond to this. In general I think that docker images should reside in [Docker Hub](https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=mailtrain&starCount=0) as there are so many different ways to set up a container.

djalal:
It's not about where it should reside, but who pushed it. Because when searching on Docker Hub, I wasn't sure or confident enough in running "non official" images. So I wondered: 
 - is there an official repository on Docker Hub ?
 - is there a simple support for containers in mailtrain ?

Pushing this from the official mailtrain Github repository is a good thing to have, IMHO.

PR #94, opened 2016-09-16
Add Cloudron installation

gramakri:
Hi @andris9,
Thanks for Mailtrain! We have been working on a Mailtrain app for the Cloudron for a while now. Would be great to have it listed as one of the installation methods.

Hopefully this gives some context on why we contributed the LDAP backend - https://github.com/andris9/mailtrain/pull/85. Users of the cloudron authenticate with Mailtrain using LDAP.

You can also play around with the app at https://my-demo.cloudron.me (username: cloudron password: cloudron).

cc @nebulade


andris9:
Thanks, looks great!

PR #109, opened 2016-11-25
Remove unused csv-generate

nebulade:
This came up as `npm test` complained about unused variable.


andris9:
I plan to use it for exporting data. Haven't got time so far to actually implement it.

PR #110, opened 2016-11-25
Add support for reply-to field

### LINKED ISSUE #108: Support specifying the reply-to field
Hi there,

is it somehow possible to set the reply-to field for a campaign? I was unable to find any information about that. 

If it is not possible, is there a specific reason, this is not implemented? If not, I am happy to work on a change adding this feature.

Thanks

nebulade:
Fixes #108 

Since this is a bigger pull request, I assume there will be some feedback.
I might have missed some places which also need adjustments, however it works in my manual testing.

Thanks for the well written code, it was quite fun to add it due to the clean code base.

Please let me know what needs to be changed.

### DIFF
--- a/lib/models/campaigns.js
+++ b/lib/models/campaigns.js
@@ -14,7 +14,7 @@ let mailer = require('../mailer');
 let caches = require('../caches');
 let humanize = require('humanize');
 
-let allowedKeys = ['description', 'from', 'address', 'subject', 'template', 'source_url', 'list', 'segment', 'html', 'text', 'tracking_disabled'];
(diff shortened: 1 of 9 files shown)


andris9:
Thanks! I'll review this tomorrow. From a quick overview it seems you should also bump the schema version on [meta.json](https://github.com/andris9/mailtrain/blob/master/meta.json) to 20.

I dont' have this documented anywhere but instead of modifying mailtrain.sql by hand you should run `npm run sqlgen` (destructive call, flushes all your Mailtrain tables!) that would apply all upgrade scripts to base.sql and dump the result as the new mailtrain.sql.

nebulade:
Thanks for the hint, about the `npm run sqlgen` I have accidentally found that while looking for how the db migration scripts work. From what I can tell this is already included in the pull request.

PR #145, opened 2017-03-04
Merge tag reference abstraction/partial

witzig:
Perhaps a news file helpers.js would be a more suitable place for getDefaultMergeTags() and getListMergeTags() … ?


andris9:
Seems reasonable

PR #158, opened 2017-03-08
Add List-Unsubscribe header

heroic:
This is useful in reducing spam complaints.


andris9:
Could you re-submit this PR without the CSRF and Template file changes?

heroic:
Done.

PR #161, opened 2017-03-08
Ses sns patch

### LINKED ISSUE #160: SES SNS notifications are not working
Ok, so I have debugged this to a great extent. It seems that the notification is delivered to the webhook, but the webhook is unable to find the response in the DB. When I check manually the response is comfortably present in the db. Now I can only assume that this could have something to do with the sender now saving the response in time. But it's not a really easy a task to find out!

Any help is appreciated!

heroic:
Fixes #160


routes/subscription.js
```diff
@@ -210,7 +210,7 @@ router.get('/:cid/unsubscribe-notice', (req, res, next) => {
     });
 });
 
-router.post('/:cid/subscribe', passport.parseForm, passport.csrfProtection, (req, res, next) => {
```
andris9: Is there a reason for skipping CSRF check?

routes/webhooks.js
```diff
@@ -41,7 +41,7 @@ router.post('/aws', (req, res, next) => {
                 }
 
                 if (req.body.Message.mail && req.body.Message.mail.messageId) {
-                    campaigns.findMailByResponse(req.body.Message.mail.messageId, (err, message) => {
+                    campaigns.findMailByResponse('<'+req.body.Message.mail.messageId+'@email.amazonses.com>', (err, message) => {
```
andris9: This is valid issue but I don't think it can be solved like this as the domain part after @ changes based on region. I'll better fix this in Nodemailer so it would return the original response instead of just the final message-id

heroic:
Are you sure the domain part changes? Mine doesn't seem to have any region
mentioned, so can't be sure.

PR #163, opened 2017-03-08
Fix the bug for SES and SNS

### LINKED ISSUE #160: SES SNS notifications are not working

heroic:
Fixes #160

### DIFF
--- a/routes/webhooks.js
+++ b/routes/webhooks.js
@@ -41,7 +41,7 @@ router.post('/aws', (req, res, next) => {
                 }
 
                 if (req.body.Message.mail && req.body.Message.mail.messageId) {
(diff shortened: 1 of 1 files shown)


andris9:
I fixed this by upgrading Nodemailer. This way Mailtrain would store not the message-id header but the raw uuid string that can be matched against the bounce notification. If you'd upgrade Mailtrain and reinstall dependencies then it should start working

Issue #156, opened 2017-03-08
Feature Request: Custom subscribe pages/e-mails for different lists


bensulli:
(Sorry if this isn't where feature requests go)

I'd like to be able to configure the look, feel and text of the signup page and other predetermined pages/e-mails on a per-list basis. I have one mailing list that's a regular mailing list that the generic template works fine for, but I also have a mailing list that serves more as an application form. I'd like this to have different text and imagery.

Currently there's only one .hbs file for the header, footer, etc, that applies to all lists. It'd be nice to have custom content in the /views folder files per list.

For now I guess I can uses PHP logic to show different content depending on the context?

andris9:
For now you could create your own forms and post subscription data to your Mailtrain installation using the Mailtrain API. You can see the API docs if you log in to your Mailtrain.

bensulli:
That's a good solution, though I don't personally have the skills to build it.

andris9:
Mailtrain now supports all this. You can create custom forms, subscription emails etc.

PR #168, opened 2017-03-10
Inject custom CSS and JS files in layout.hbs

witzig:
Also adresses #156.


witzig:
Perhaps the changes to config/default.toml could be dropped and a note in README.md would be sufficient. Yet getting this into master would probably prevent a few forks for production. At least in my case. 😉

andris9:
Having the explanation if default config is OK as its following similar style as with other config options

PR #172, opened 2017-03-11
Some translation and layout fixes 

witzig:


views/partials/html-to-text.hbs
```diff
@@ -1,7 +1,7 @@
 <div class="form-group">
     <div class="col-sm-offset-2 col-sm-10">
         <span class="help-block">
-            {{#translate}}To extract the text from HTML click <a id="html-to-plaintext-btn" role="button">here</a>.{{/translate}}
+            {{#translate}}To extract the text from HTML click{{/translate}} <a id="html-to-plaintext-btn" role="button">{{#translate}}here{{/translate}}</a>.
```
andris9: I'm not sure about this one. Having some HTML code in the translation string is not an issue. If sentences are split then this might cause strange results with some languages.

PR #174, opened 2017-03-11
Fixed some translation tags

witzig:


witzig:
New PR without the change partial/html-to-text.hbs

PR #178, opened 2017-03-15
[bugfix] Fix LDAP issue with OpenLDAP/MS AD

larrabee:
LDAP not worked if "username" field not present. I was add new parameter to config to change username field name if needed.

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -112,6 +112,8 @@ host="localhost"
 port=3002
 baseDN="ou=users,dc=company"
 filter="(|(username={{username}})(mail={{username}}))"
+#Username field in LDAP (uid/cn/username)
+uidTag="username"
 passwordresetlink=""
 
 [postfixbounce]
--- a/lib/passport.js
+++ b/lib/passport.js
@@ -16,8 +16,9 @@ let LdapStrategy;
 try {
     LdapStrategy = require('passport-ldapjs').Strategy; // eslint-disable-line global-require
 } catch (E) {
-    // ignore
-    console.warn('Module "passport-ldapjs" not installed. LDAP auth will fail.');
+    if (config.ldap.enabled) {
+      log.info('LDAP', 'Module "passport-ldapjs" not installed. LDAP auth will fail.');
+    }
 }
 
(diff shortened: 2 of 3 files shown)


andris9:
Maybe @nebulade can review this? I don't know much about LDAP.

nebulade:
I will look into it. Already now I can tell that the `package.json` should not include the ldapjs passport module. At least I had it initially also included in my first PR, which was correctly mentioned to not be needed.

larrabee:
Hello. Without passport-ldapjs module auth fails with error "Unknown auth type LDAP".

nebulade:
I think the intention was that if you configure to use ldap then you manually install that module. But since most people don't use it, no need to pull in the dependency by default. Maybe @andris9 can give his opinion on that as well.

nebulade:
The change otherwise looks good. Maybe I can research a bit if there is a common pattern at other popular project's ldap integration to name the username field `uidTag`

andris9:
@nebulade thanks! I did not want to blindly accept this as it might have caused issues with existing LDAP users.

The problem with passport-ldapjs was that it includes a compiled dependency (dtrace_provider) that might not work everywhere. It seems though that it is an optional dependency, not a strict requirement, so I guess that it would be OK to include passport-ldapjs in Mailtrain dependencies – in worst case you get a long list of compilation errors but it should not break the Mailtrain installation process.

larrabee:
Ok, that about [this](https://github.com/andris9/mailtrain/pull/178/commits/1ca27b6b404e6e4be181be846c5a97dc9e61b190)?

andris9:
@larrabee I think that the warning should be shown only if the LDAP authentication is actually enabled. You can check it from `config.ldap.enabled`.

Instead of `console.warn` use the included log module:

    log.info('LDAP', 'Module "passport-ldapjs" not installed. LDAP auth will fail.');

andris9:
Thanks!

larrabee:
Thank you ;)

nebulade:
Great to see more usage of the ldap part! I only really put in just enough effort to get it working well on Cloudron, instead of making it very generic, like it is now a bit more.

PR #179, opened 2017-03-17
[feature] Add Dockerfile and update readme

larrabee:
Hello. This pull request add dockerfile and install instructions to readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,9 +157,12 @@ With proper SPF, DKIM and PTR records (DMARC wouldn't hurt either) I got perfect
         - "3000:3000"
       volumes:
         - "/etc/mailtrain/production.toml:/app/config/production.toml"
+        - "mailtrain-node-data:/app/public/grapejs/uploads"
+        - "mailtrain-node-data:/app/public/mosaico/uploads"
   volumes:
     mailtrain-mysq-data: {}
     mailtrain-redis-data: {}
+    mailtrain-node-data: {}
 
   ```
   5. Update MySQL and Redis credintial in `/etc/mailtrain/production.toml` like this:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -115,6 +115,77 @@ If you are using the bundled ZoneMTA then you should make sure you are using a p
...
+      image: mailtrain-node:latest
+      container_name: "mailtrain-node"
+      links:
+        - "mailtrain-mysql:mailtrain-mysql"
+        - "mailtrain-redis:mailtrain-redis"
+      ports:
+        - "3000:3000"
+      volumes:
+        - "/etc/mailtrain/production.toml:/app/config/production.toml"
+  volumes:
```
witzig: What about GrapeJS and Mosaico upload folders?

larrabee: Thanks, did not know about it.

andris9:
Dockerfile support has been requested before. Usually I have not been in favor of adding it as Dockerfiles can be composed in many ways and different users might prefer different setup (eg. by using a specific base image)

larrabee:
Many users like install app by docker pull/build && copy-paste config && run. With default Dockerfile they can do it.  
Other user still able to build own docker image or change existing dockerfile.

andris9:
OK, I guess it would make sense to add it. I think you should also install imagemagick, otherwise Mosaico image uploads would not work properly.

witzig:
Yeah, a Dockerfile was the first thing I was looking for when discovering Mailtrain. Would be be great to have this.

osixia:
Is there any reason to use centos as baseimage instead of more light weight baseimage like the official node:7.7.4-alpine for example ?

PR #180, opened 2017-03-19
Custom forms mjml

witzig:
Take your time ... ;)


andris9:
Could you explain a bit the barracuda change? Changing database options might not be an option in a lot of cases, where you do not manage the database yourself but host it from a provider.

witzig:
Yeah, I was thinking in the context of the dedicated VPS and picked the low hanging fruit.

InnoDB file format Antelope only supports row format COMPACT and REDUNDANT, storing the first 768 bytes of variable length columns (BLOB, VARCHAR, TEXT) in the row, the remainder in dedicated pages. Barracuda also supports row formats COMPRESSED and DYNAMIC, storing a 20-byte pointer in the index. To my knowledge, the default InnoDB file format in MySQL 5.6 is Antelope to ensure maximum backward compatibility and has been changed in MySQL 5.7 to Barracuda.

Thus, with Antelope using row format COMPACT, the newly created table custom_forms hits the row size limit of 8126 bytes (at least in my case).

Running dbcheck.js during service start and creating a table with row format DYNAMIC when file format is Antelope may silently create the table as COMPACT (https://bugs.mysql.com/bug.php?id=83647) or throw an error, preventing Mailtrain to start. To avoid this downtime, I assumed trying to set the file format to Barracuda would be in the interest of most users, at least in the context of the dedicated VPS. But like I wrote in mail, it doesn't feel right.

Perhaps just checking for Barracuda, logging instructions, and letting errors be errors would do. 

Using MyISAM instead of InnoDB for table custom_forms would be an alternative, at the price of losing foreign key constraints.

Generally, I reckon the way forward is InnoDB and Barracuda.

andris9:
Ok, I did not know about that limit. Anyhow, in my own cases I use shared MySQL databases and the mailtrain user has only limited privileges and I assume this also applies to people using Amazon RDS and similar.

Wouldn't it make more sense to break this table up to get past that limitation? Something like custom_forms for the metadata but all text content would reside in custom_forms_data

`custom_forms`:

| id | list | name | ... |
|---:|---:|----|---|
| 100 | 200 | custom form | ... |

`custom_forms_data`:

| id | form | key | value |
|---:|---:|----|---|
| 1 | 100 | web_manage | `'<p>template</p>'` |

I would also change TEXT to LONGTEXT, otherwise it would break if you try to use HTML with data-url images

witzig:
> Wouldn't it make more sense to break this table up to get past that limitation?

Done.

Although in hindsight I'm not sure if it'd been wiser to got with MyISAM, implement delete cascade ourself, and change the table to InnoDB down the road.

andris9:
Seems great! The only thing I'd do is to add some kind of note about MJML usage (there's the Help tab but I don't think it is too easy to spot). Maybe a note between the dropdown and code editor, something in the lines of "custom forms use MJML for formatting, see the documentation here" or something like that.

witzig:
Will be added.

Also dropping the references from table custom_forms to custom_forms_data. Thought it might come in handy to have their ID's for lazy loading, but it's redundant.

PR #187, opened 2017-03-31
Add option to disable email cheking for imported CSV

larrabee:
Hello. This request add option, that can disable email checking for emails, imported from CSV file. This option useful for speedup import big csv (1+ million rows) files.
[Screenshot](https://ibb.co/hOoo8v)

### DIFF
--- a/lib/models/subscriptions.js
+++ b/lib/models/subscriptions.js
@@ -982,7 +982,7 @@ module.exports.delete = (listId, cid, callback) => {
     });
 };
 
-module.exports.createImport = (listId, type, path, size, delimiter, mapping, callback) => {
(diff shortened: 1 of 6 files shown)


andris9:
Thanks! It is definitely faster that way, as the current mail validation logic is really suboptimal.

I recently wrote a list validation service for a client where I was able to achieve about on average 1000 address validations/second on a cheap VPS and this included format validation, DNS resolving and even connections to MX to see if the address is accepted by the server or not.

As the code was done for a client then I can't re-use it for Mailtrain but if I ever have time I try to come up with something similar. For now, I guess, just skipping DNS checks is already a good enough option, especially if you have already validated your list.

Issue #194, opened 2017-04-07
Dockerfile - user local files rather than cloning latest


programster:
I just downloaded release version 1.23 (which happens to currently be the latest release, but may not be in future). I ran through the documented docker build steps, but then noticed the Dockerfile was just ignoring the version I had downloaded and always fetches the latest code with the step below:

```
git clone git://github.com/Mailtrain-org/mailtrain.git /app
```

Please could we have the dockerfile use the local files instead, so that if a user wishes to deploy a specific release, they can just download that version, and run a docker build from there. A quick hack to get this to work would be to move all the files (except the Dockerfile) into a folder called "app" and replace the command: 

```
git clone git://github.com/Mailtrain-org/mailtrain.git /app
```
with 

```
COPY app /app
```

andris9:
Do you have suggestions how to best handle this so it would not get out of sync? Using some git hook that automatically checks that the version in Dockerfile matches package.json?

programster:
@andris9 I'm a bit confused as to what has to be in-sync. Are you talking about the name/tag of the docker image that get's built when you run a docker build or perhaps libraries that are included at the last minute? My confusion is probably because I'm coming from a PHP background rather than a node one and not sure what that package.json is for but can only assume its like composer.json for including external packages/libs. Ideally (at least from my POV) "releases" are bundled with any/all necessary code with them. E.g. in PHP based applications, the releases would have the "vendor" file included in them, whereas the "master" branch in git would not include vendor and it would be up to the developer to run "composer update". Alternatively, one can use "fixed" version numbers of packages.

andris9:
By keeping things in sync I'm mean that for now releasing a new Mailtrain version means tagging current state in master branch. The Dockerfile was provided by another contributor and sooner or later it is forgotten to update it. So let's say I change the Dockerfile and set the version to use as v1.23.2. Next month the latest version could be already 1.25.0 but the Dockerfile is still at 1.23.2 as no-one remembered to update it.

witzig:
Wouldn't it be possible to fetch the tags here (https://github.com/Mailtrain-org/mailtrain/blob/master/Dockerfile#L4) too, then using build args (https://docs.docker.com/engine/reference/builder/#arg) to checkout a specific version?

programster:
@andris9 If you implement the change I suggested of changing the line in the Dockerfile from:

```
git clone git://github.com/Mailtrain-org/mailtrain.git /app
```
to
```
COPY app /app
```
[...]

PR #199, opened 2017-04-08
Docker changes for using local files rather than pulling from web

programster:
Using the local files for the docker build rather than performing a git clone. This way if we build from an older release, we will be building that release rather than always the latest version of mailtrain.

This pull request is a one line change that does not require moving any files around.

Refer to the discussion [here](https://github.com/Mailtrain-org/mailtrain/issues/194).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM centos
 RUN curl --silent --location https://rpm.nodesource.com/setup_7.x | bash -
 RUN yum install -y git make gcc nodejs ImageMagick && yum clean all
-RUN git clone git://github.com/Mailtrain-org/mailtrain.git /app
+COPY . /app
 WORKDIR /app/
 ENV NODE_ENV production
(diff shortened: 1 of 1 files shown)


andris9:
This one seems better. Should the README for Dockerfile usage be updated as well?

programster:
@andris9 I just re-read through the Dockerfile and it **doesn't** appear to me like the README needs to be updated at all in any way. The only thing that could possibly need looking at is the part that says to:

> Run sudo docker build -t mailtrain-node:latest .

Perhaps this should just be changed to somehow tell the user to only use `latest` if building from mater, and use the version number if they are building from a release. To be honest, if a user is building from source they shouldn't really care about getting the right tag and can just have latest. It is only the owner/maintainer of the official dockerhub image that should care. Correct me if I'm wrong, but I can't find an official image which would be nice to have one day.

andris9:
Ok, thanks

Issue #2, opened 2016-04-05
PostgreSQL Support


2silver:
As alternative for MySQL

andris9:
Postgres support is not something I'm going to do myself but there's a guy who offered to help with Postgres support whenever he has time for it (which might mean tomorrow or never, who knows)

michaelBenin:
What about using an orm? Something like bookshelf?

andris9:
I'm not sure how well an ORM would perform in this case – I use a lot of foreign keys, creating and deleting tables (every list and campaign gets its own tables to support large lists and to track individual clicks), adding and removing columns/indexes to tables (needed to support custom fields), transactions (bounced message should also decrement subscribers counter without going out of sync etc.) and so on.

michaelBenin:
From my experience, what you described can be accomplished with Knex and Bookshelf.

michaelBenin:
Also, it would allow flexibility with other supported dbs.

forgeOneHQ:
I am closing this issue again.

saper:
Most of the work needs to be done in the https://github.com/smartarch/ivis-core dependency, which, in turn, uses [Knex](https://www.npmjs.com/package/knex), so it might be possible to build something working on PostgreSQL.

PR #207, opened 2017-04-14
Protected lists

bures:
I started adding a few features to Mailtrain that we might need to use it for distributing newsletters and other messages. Here are the first ones:

- Added option to mark a list as not being allowed to be subscribed by public users using the form. The settings is a checkbox in list create/edit.

- Co-deduplication in list and filter methods for ajax tables. I would like to write two table-based views and didn't want to just copy paste the existing code. This should make it easier now.


bures:
Here is another piece of functionality. It allows defining and generating reports. Essentially the idea is the following:

An administrator creates a report template (there are some examples packaged with the feature). The report template defines what formal parameters is needs (selection of campaigns, lists, etc.), the JS code for processing the data and hbs template for rendering them.

A user creates a new report based on an existing template. In this process he/she defines the actual parameters (e.g. a particular campaign or list). Mailtrain generates the report. The generation of reports happens asynchronously in a dedicated worker process. The user can view it or download it. Optionally the user can re-run the report generation, stop to generation, view outputs of the JS data processing code (via console.log).

When creating the report template, it is possible to choose between mime types text/html and text/csv. If it is text/csv, Mailtrain allows downloading the "report" instead of viewing it. This assumes that the hbs template simply generates the CSV format. This is particularly useful for exporting various stuff (e.g. lists). I noticed that this is functionality which has been asked for multiple times.

I'm attaching a few screenshots.

Please note that I did test it, but not very thoroughly. There may still be some rough edges.

BTW, I'm not sure what your process it. The previous pull request was approved by Roger, but was never merged. As far as I understand privileges in GitHub, I can't merge it myself. It has to be done by someone with write privileges. Or did I misunderstand anything?

Thanks.
Tomas

edit-report-template-2
reports-overview
report-templates-overview
view-report
create-list-export
create-new-report
edit-report-template-1

witzig:
> BTW, I'm not sure what your process it. The previous pull request was approved by Roger, but was never merged.

Merging is the responsibility of Andris. I'm sure you'll get a feedback from him shortly, perhaps over the weekend. It's a huge PR, so it might take a bit longer. In the meantime It'd be great if you could run `grunt` from the command line and fix the ESLint errors.

There're some further commits since I last reviewed, namely:

> Note that the execution of custom JavaScript in the data retrieval / processing code and definition of custom Handlebars templates is a security issue.

I'm not sure about the execution of custom JavaScript, yet I reckon it's the most flexible and convenient solution ...

> Please note that I did test it, but not very thoroughly. There may still be some rough edges.

@andris9, I tested it too. Works lovely.

andris9:
Hey, I'm currently occupied with [Locked Shields exercises](https://ccdcoe.org/largest-international-technical-cyber-defence-exercise-world-takes-place-next-week.html) where I'm in one of the blue teams, so I dont have much time on reviewing stuff. @witzig feel free to merge anything that you're fine with, I'll tag official releases sometime later.

bures:
Perfect. Thanks. I'll take a look at the ESLint errors and let you know once this is fixed.

As for the the execution of custom JavaScript. I know it's a security issue. I first wanted to do it with solely with handlebars, but then I required to do various aggregation and computation, which was awkward to do in handlebars. In the end, I concluded that the custom JS might be an easy way around it. 

Generally, it should be possible to sandbox the code. It does not really need to require anything save for the modules from "lib/model" to be able to query the DB. If we had some safe query abstraction over the database we might be able isolate it. What do you think? 

One of my future plans to is add proper user management with granular permissions. If this is in place, the definition of a template should be restricted to an admin.

BTW, have you considered any ORM to abstract the entities in the DB? I tried to deduplicate the code a bit, but still, there is still that could be done in "lib/models".

witzig:
@andris9, thanks, and good luck!

Tomas, concerning the execution of custom JavaScript (as is), I'm a bit hesitant to merge and would prefer to have this reviewed by Andris. The sandboxed solution you suggested would be preferable, IMHO. 

Another thought, although not as flexible, but perhaps 95% of the use cases could be covered with predefined 'data retrieval functions', and a dedicated folder containing custom snippets? Yet this might just be a shortsighted shortcut (around the security issue) on my part.

> One of my future plans to is add proper user management with granular permissions. [...]

Oh, that would be fantastic and widely appreciated/anticipated!

> BTW, have you considered any ORM to abstract the entities in the DB? I tried to deduplicate the code a bit, but still, there is still that could be done in "lib/models".

Yeah, there's a lot of deduplication potential. :) As for ORM abstraction see Andris' [comment from April 2016](https://github.com/Mailtrain-org/mailtrain/issues/2#issuecomment-205867427). I guess we could start working on this on a feature branch and see how it fares.

bures:
I think the sandboxed solution with some data retrieval should be pretty easy. It is already sandboxed, only I now expose "require" to the sandbox, which allows the code to escape it. 

I would expose to the JS sandbox a few "safe" data retrieval functions for campaigns and lists and the "console" object (for logging).

Would you be OK then with running the JS code in such a sandbox?

As for ORM, I like the idea of Knex and Bookshelf. Shall we start a dedicated issue on it to discuss it thoroughly? I don't think it should be discussed under "PostgreSQL Support".

witzig:
> Would you be OK then with running the JS code in such a sandbox?

As far as I'm concerned, yes. I can't think of an argument against it if the code can solely read within its scope.

As for ORM, please feel free to open an issue for discussion. I too like the idea of Knex and Bookshelf and think it could be worth the effort. However, I guess it'd be foolish to assume that Andris didn't have good reason to skip ORM in the first place ... ?

bures:
The eslint errors are fixed now.

I looked in the sandboxing options. It seems that I was originally too optimistic. The only somewhat reasonable solution is vm2 (https://github.com/patriksimek/vm2), but even that looks like a futile effort to patch all the different paths one can sneak through in a prototype based language.

In my view the only solid way of isolating anything running nodejs is on the OS level. Which brings me to the question:

@witzig What kind of security do you have in mind? Not being able to overwrite files? Not being able to modify DB? Or anything more? 

Thank you.

witzig:
@bures, for clarification, I'm just the editor guy. Like yourself I needed some features Mailtrain was missing. As a result Mailtrain has since been converted into an organisation whereby my scope is to maintain these features (and helping a bit out with tickets and reviewing stuff).

Hence this question is actually out of my domain and can only be answered by @andris9. For what it's worth or IMHO, (over)writing files is out of the question. Executing arbitrary SQL should not be possible as well. And not being able to read stuff (ie. configs) would be good too.

bures:
Sounds good. What about the following?

- the worker process will chroot to protected/reports and drop privileges before it executes the user JS

- we will have one more DB user (mailtrain-ro), which will have only read-only access to the DB; the configuration loaded in the worker process will contain the mailtrain-ro DB user (instead of the regular one)

That should give us those two properties: no modification of files, no modification of DB.

Note that this still does not prevent the one creating a custom template to perform XSS. However, I believe this should not be a problem, because creating/editing report templates should be restricted solely to admins, which have access to anything in the web interface anyway. Thus they won't benefit from doing XSS.

@witzig What do you think?

@andris9 It seems we would really need at least some indication of your preferences. If this could be somewhat soon, I could do the modifications in the upcoming days. Thanks.

andris9:
@bures my only limitation regarding Mailtrain is that it should work on these webhosting providers that have Node installed in their server next to the usual LAMP stack. This means that Mailtrain should be able to start in non-root mode and be able to use a single limited MySQL user with a single database. If you append on that, eg. there's an _option_ to use a more secure second MySQL user but it's _not compulsory_, then it's fine by me.

bures:
Sounds good.

So I would suggest the following:

- The reporting feature is optional - can be disabled in config if the risk of running user defined JS seems too high.

- The chroot and read-only DB user is again optional - can be enabled in config if one wants to play safer.

What do you think?

andris9:
Sounds great!

bures:
Perfect. Thank you both. I'll come back once it is done (or if I figure out any other obstacle :-) ).

witzig:
Perhaps, if we make the reporting feature optional, the chroot and read-only DB user should be omitted for the sake of simplicity. The PR would technically be good to merge that way (if I read correctly).

bures:
I improved the security a bit. It prevents write access from the custom javascript and allows it only read access to DB. 

So here a summary:

Reports functionality is disabled by default. Can be enabled by config.reports.enabled = true.

Reports run under special user configured in config.rouser and config.rogroup. If they are left undefined, Mailtrain uses config.user/group or the current user if config.user/group are not defined.

Reports worker has separate config (under workers/reports/config). This config can specify different mysql user/password. The idea is that this user has only privileges for SELECT. 

The existing config directory has to be made inaccessible to "other". This prevents the rouser to read DB password for RW access to the DB.

Install scripts have been extended to create the ro user (system-level and for the database). There is additionally an install script for CentOS 7 (my primary platform)

The bug with the senders running as root is fixed. The privileges are now dropped much sooner.


Technically a refactored a few classes. The most important thing from the architectural perspective is the Executor (services/executor.js). It is a separate process that retains root privileges even after Mailtrain drops them. Mailtrain communicates with the Executor via child_process.send. The executor spawns new report workers and then reports back to Mailtrain if the finish, etc.
The executor is intentionally written in rather minimalistinc manner (it does not include much custom functionality). This should make it relatively secure to keep the root privileges.

In the future, I was even thinking that the executor could create sockets < 1000 and pass them to children. This way, we could switch Executor and current index.js. The Executor would start Mailtrain main app with restricted privileges and pass it sockets for VERP and for HTTP. This would mean that we don't as root with all battery of custom modules transitively included in index.js. If you like this extension, I can implement it once we finish with this pull request.

witzig:
@bures, thank you so much for your work, this PR looks great all around! As Andris' Locked Shields exercises have ended today, I think we should give him the weekend to take a final peek.

@andris9, I tested again, works great!

andris9:
Awesome!

andris9:
PS. @bures I invited you to be a member of Mailtrain-org so you would be able to change stuff if there's something that needs to be updated regarding the added features

andris9:
Oh, and yes, Locked Shields is now over and we did quite well (second place) but got beaten by the Czech team. https://ccdcoe.org/czech-team-wins-cyber-defence-exercise-locked-shields-2017.html

witzig:
@bures, also from my side, it would be awesome to have you on board! And to be frank, you should rather be reviewing my stuff than the other way around. :-)

@andris9, congratulations! We expected nothing less of you. :-)

bures:
Thank you for the invitation. I'm happy to be onboard.

@andris9 - Congratulations to the second place. Of course, it's nice to hear too that it was the team from my country that won. :-)

rikmendes:
I'm so amazed by this PR & overall feature, trying list export to CSV now :)

Issue #106, opened 2016-11-17
No emails are being sent.


stevenmcastano:
I've got a list, template, campaign, and when I enter my SMTP settings everything verifies and says I'm ready to send mail.

However, when I click to start sending a campaign, I see the spinning arrows saying it's sending, but it never makes and progress forward and looking at the logs on me email server, it's not even attempting to send the emails. Not a single connection.

I have the box checked to log SMTP transactions also, but I can't seem to find the log or anything that might be wrong with it.

What steps should I take next to troubleshoot?

andris9:
Do the subscription confirmation emails work?

stevenmcastano:
Yes, they do and based on the email headers, they're being sent through the proper server via SSL, all my DKIM and SPF records are in place, everything looks good to go. I tried subscribing with emails from gmail, yahoo, and two of my own custom domains.

stevenmcastano:
Out of nowhere a test campaign I left trying to send HOURS ago to a list with just my email address on it just went through without even doing anything... it's just been sitting there saying it was sending without doing anything for hours.

Any ideas?

stevenmcastano:
Now the other campaign to the original 30 or so users just all of a sudden went out, and the odd part is that I set it to sent right after the last single email went out, which is when I posted here last... which was 5 hours ago and looking at the list of delivered emails... they all say they were sent/delivered 5 hours ago as well in mailtrain, but the log on the server as well as two of my own email addresses on the list all show the emails were just sent a few minutes ago. So it seems like they're on some sort of almost exact 5 hour delay.

Is there someone with time zones? delays? other time configuration? etc....?

stevenmcastano:
Even users clicks that are happening right now as they're getting the emails are still being recorded as having happened "5 hours ago"

andris9:
Yes, seems like a timezone issue. Could you try modifying the mysql config and change timezone from "local" to "-0500" or "0500" to see if the shift disappears?

stevenmcastano:
Looks like the timezone remove the shift in times and it looks like it got emails to send ASAP... I only tried it to my test list of one users, but this time when it hit send it prepped the email and sent right away. I'll have to give it some time until I have another excuse to both my customers.

Any idea why the timezone setting would actually delay email sending??

andris9:
[...]

PR #216, opened 2017-04-28
More verbose upgrade instructions

christiangenco:


witzig:
@christiangenco, thanks for the PR and the [detailed solution](https://github.com/Mailtrain-org/mailtrain/issues/106#issuecomment-298086175) to your timezone issue. 

However, the proposed upgrade instructions are quite specific and don't apply to many installations, e.g. `cd /opt/mailtrain` or `service mailtrain restart`.

Also `/etc/init.d/mysql restart`, while solving your particular issue, isn't usually required. To get to the bottom of this, did you really just ...

> I just did a cd /opt/mailtrain; git pull and am on ec7699. After upgrading I did npm install --production; service mailtrain restart.

... without any other server modifications, i.e. upgrading mysql or alike? And which version of Mailtrain did you run before the upgrade?

Thanks.

Issue #222, opened 2017-04-30
A subscriber address cannot be changed to an address which has been previously unsubscribed


bures:
I did not test this, but it seems from the code (lib/models/subscriptions.js:1190) that if an address cannot be changed to an unsubscribed one. 

In order to do so, the SQL statement should additionally contain something as "AND status=1".

Please let me know if this I misunderstood anything. Otherwise, I'll fix it as part of the extension of the subscription mechanism.

witzig:
The email address is a unique index of the subscription tables, so I guess we would have to delete the 'old' membership upon confirmation, then change the address of the 'current' membership.

andris9:
Yes, the old subscription needs to be removed in this case or alternatively changed to something else (_user@address_ -> _removed.randomstring.user@address_) to keep the references in old campaigns.

bures:
It's an interesting idea to rename it to something else. So what is the consensus. Myself, I would vote for removing it. The reasoning here is that if someone changes to address to an unsubscribed entry, then it must be the same guy as before and he/she essentially subscribed again. So, the same way we don't keep history of all unsubscriptions/resubscriptions, we would not keep the history here too.

andris9:
Well, yeah, I guess the old entry can be deleted. There's already a [delete method](https://github.com/Mailtrain-org/mailtrain/blob/master/lib/models/subscriptions.js#L818) for it.

witzig:
Fixed with #222. An e2e test has been added with #227.

Issue #221, opened 2017-04-30
Anonymous user can unsubscribe pretty much anyone if he/she knows the email address


bures:
I'm not sure whether this is a bug or a feature. It seems that an anonymous user can unsubscribe pretty much anyone whose email he/she knows. This is because the unsubscribe form allows filling in an email which is not checked against the email of the one who clicked the link.

Steps to reproduce:

1) subscribe to a list
2) click unsubscribe in an email
3) fill in an address of someone else and click unsubscribe

I think we should change this to make it possible to unsubscribe only the user which the unsubscribe link belongs to.

If OK, I will include the change in the extension of the subscription/unsubscription process that I'm currently working on.

witzig:
Since the unsubscribe form is only accessible with a valid `ucid`, that's definitely a bug. Thanks in advance for the patch.

andris9:
It is hard to say if it's a bug or not because I think this is what you can do with other services as well. Usually you don't need special permissions to unsubscribe, just enter an email address and it's gone.

witzig:
Ok, let's not call this a bug per se, yet I'd still argue it shouldn't be possible at the current URL `/:lcid/unsubscribe/:ucid` which is tied to a specific subscription. Perhaps a generic unsubscribe form at  `/:lcid/unsubscribe` would be more suitable. That URL could also safely be published, analog to the URL of the subscription form.

andris9:
Should the address input field on the unsubscribe form be replaced with a static address text then? If you can only unsubscribe yourself then the input field doesn't make much sense anymore. I'd be fine with such a change.

bures:
Yeah. Actually, I already replaced it with static text in my development version. So we are on the same page.

Issue #220, opened 2017-04-30
Manage address feature of the subscription functionality leaks out membership in a list


bures:
If someone tries to change an address to another which is in the list, the system reports that the new address is already registered. This is not a bug, it's a feature. However, it would be nice if the system does not give out this information.

I would suggest that it would issue some generic message as: "Further instructions have been sent to the new email." Then it would either send an email with the confirmation link (like it is now) or an email saying that someone tried to register the email, but the email is already registered.

The same (or similar) message about email already registered could be issued when someone tries to subscribe to the list with an email that is already subscribed. This would be part of the approach suggested by Roger in #218.

If OK, I'll implement it. Anyway, I'm currently refactoring a bit the subscription funcionality - removing some duplicities and implementing #219.

witzig:
That seems a very sensible suggestion to me. Not leaking memberships when subscribing #218, managing, or unsubscribing #221 could be an essential requirement for some lists.

andris9:
Sounds reasonable

Issue #219, opened 2017-04-30
Selectable unsubscription process


bures:
We plan to use Mailtrain for public newsletters and also for newsletters inside a non-profit organisation. Our subscribers are far from tech savvy and sometimes they forward the newsletters directly from their mailboxes.

All this leads to several issues:
1) If a subscriber forwards the newsletter on his/her own, they forward it together with the unsubscription link, which typically results into someone else unsubscribing them.

2) The internal newsletters inside the organisation should not be subscribable by public (this is already handled by the pull request #207) and they should not allow direct unsubscription. Rather the unsubscription should go through a superior. The reasoning here is that knowing the information contained in the newsletters is part of the responsibilities of the workers in the organisation. 

For these two cases, we would need more flexibility in how the unsubscription process works. Essentially, I see that a list would allow selection of a unsubscription process:
1) One-step unsubscribe - this is the same as it is now
2) Two-step unsubscribe - this would include confirmation email with a link that has to be clicked in order to complete the unsubscription (this addresses the case 1 above).
3) Manual unsubscribe - clicking the unsubscribe link would go to a page with information whom to contact to get removed from the list (this addresses the case 2 above).

If you are OK with this, I already started extending the subscription functionality to include the cases above. I believe this is one of the features needed in Mailtrain to be able to target more enterprise use-cases.

witzig:
That sounds reasonable to me.

andris9:
I don't have strong preference in this issue, as long as there's a way to unsubscribe it's fine by me.

What about adding an option to not auto-unscubscribe but require manual entry for the email field? So if clicked on an unsubscribe link you would end up in an unsubscribe page but it is not pre-filled. Whatever you enter into it gets unsubscribed.

witzig:
The extended subscription/unsubscription functionality is in Master now.

![Extended subscription/unsubscription](https://cloud.githubusercontent.com/assets/4999980/26661227/cbb8f886-467c-11e7-94cb-5265af2ba6f5.png)

PR #227, opened 2017-05-06
Extended subscription/unsubscription (selectable unsubscription process)

bures:
This PR represents some extensions and refactoring of the subscription/unsubscription process.

It addresses issues #218, #219, #220, #221, #222

It adds the re-subscribe feature, which was probably intended, but somehow not completed before. This means that when "Subscribe" link is clicked in unsubscribe confirmation, the form fields are pre-filled with the original values. The subscription mechanism (already existing) ensures that the subscription gets the same subscription cid as before.

Further, it contains some refactoring and code de-duplication to make future extensions of the subscription/unsubscription process easier. The most important change are:
- Sending emails moved solely to router. (Before it was partially in model and partially in router). Further, all subscription-related emails have been factored into helper methods, all in subscription-mail-helpers.js
- Confirmations functionality refactored to be generic (not specific to "subscribe"). It is now used to "subscribe", "change-address", and "unsubscribe". It can be further used for other actions that need email confirmation (even beyond the subscription/unsubscription process). For this reason, the confirmations have been moved to a dedicated model package.
- Some of the custom forms have been renamed to have more unified naming. Some more have been added to address #220, #219.

### Potential upgrade issues:
- The schema of the "data" field in the confirmation table differs to the previous version. It would be difficult to reconstruct it in the upgrade SQL-script. Thus, I decided to prune the contents of the confirmation table on schema upgrade. This means that all subscribe confirmations become invalid. The system will then give a message saying that the request is invalid and that subscription should be done again. Hopefully this solves the problem (though it will be annoying in some singular cases.)

- The PR adds a few more custom forms. If a list has some custom forms, the content of the new custom forms will be empty. It is thus necessary that after the upgrade, one goes through all existing custom forms and updates them according the the provided defaults.

### Testing:
I hope I tested the whole subscribe/manage/unsusbscribe features. I also tested the API endpoints, subscription widget and subscribe/unsubscribe in list management. However, it is possible that I forgot about some dependency or just didn't know about it. Please give it some try yourselves.

Thanks a lot. So, user management now? :-)
T.


witzig:
Awesome, thank you.

Concerning upgrade issues, I think custom forms fall back gracefully and thus this is only an issue for people who have their forms translated. Styling should be unaffected. 

> The schema of the "data" field in the confirmation table differs to the previous version. [...] Thus, I decided to prune the contents of the confirmation table on schema upgrade.

Shouldn't we keep those for legal reasons?

> I also scraped the "auto" parameter which automatically submits the unsubscription form when the link is clicked in a campaign email.

Regarding the List-Unsubscribe Header, would it be advisable to adhere to the [guidelines of the Certified Senders Alliance](https://certified-senders.eu/wp-content/uploads/CSA_list_unsubscribe.pdf)?

@andris9, I tested it too. Works great!

bures:
Thank you Roger for testing it.

>> The schema of the "data" field in the confirmation table differs to the previous version. [...] Thus, I decided to prune the contents of the confirmation table on schema upgrade.

> Shouldn't we keep those for legal reasons?

As far as I remember, the confirmations contained only pending confirmations. Once a confirmation was process, it was deleted. I preserved this functionality. Generally, the confirmations table should be rather empty and not much useful for any auditing / legal purposes.

>> I also scraped the "auto" parameter which automatically submits the unsubscription form when the link is clicked in a campaign email.

> Regarding the List-Unsubscribe Header, would it be advisable to adhere to the guidelines of the Certified Senders Alliance?

OK. Now I understand why it was there. :-) I'll reintroduce it and let you know.

witzig:
Oh, very nice! And sorry for the extra work. Not sure I should have pushed the draft in retrospect. :-)

bures:
No worries Roger. I'm very glad that you did the e2e tests. We needed to do them and this was an excellent opportunity. I also liked the way you did it. I followed your structure, just extended it a bit.

In particular, I wanted something that from the perspective of Mocha behaves as a single test, but itself is structured to steps. The abstraction of a use-case was very natural here because it describes interaction between a user and the system.

The idea here is that a Mocha test (being primarily a unit test) can generally be executed independently of other tests (provided that before and after hooks are called). From this perspective, the use-case is a single test, because one cannot just repeat the step that failed previously - instead all steps leading to it have to be repeated.

To support this better, I created a custom reporter for Mocha, which is aware of use-case steps and can display them. If another reporter is used, it interprets the whole use-case as one test and just the steps won't be displayed.

I now just would like to finish the use-case tests for all the different variants of subscription.
Once this is done, it can hopefully motivate us to create e2e tests for the new functionality.

witzig:
Oh sure, the extensions and clean up make a ton of sense. I also like your composition approach, quite elegant.

> I now just would like to finish the use-case tests for all the different variants of subscription.

Happy to help with this ...

test/e2e/tests/subscription-uc.js
```diff
@@ -0,0 +1,106 @@
...
+
+    useCase('Unsubscription from list #1 (one-step, no form)', async () => {
+        const subscription = await subscribe(generateEmail());
+
+        await step('User clicks the unsubscribe button', async () => {
+            await page.mailSubscriptionConfirmed.click('unsubscribeLink');
+        });
+
+        await step('System show a notice that confirms unsubscription', async () => {
+            await page.webUnsubscribedNotice.waitUntilVisible();
```
witzig: This works fine with `phantomjs` yet fails when using `firefox` as the `elementToWaitFor` (homepageButton) is already present in `mailSubscriptionConfirmed`. Sleeping a second helps, obviously, but perhaps a better approach might be to something like `textToWaitFor`?

bures: Roger, thanks a lot for doing this interim testing. Really greatly appreciated. I suspected this could be an issue.

Well, one "correct" solution that comes to my mind would be to extend waitUntilVisible with a piece of custom browser code something like document.mailTrainRefreshAcknowledged=true.

Then we could have a new function waitUntilVisibleAfterRefresh which would first wait for the document to lose the mailTraiRefreshAcknowledged property. Only then it would start waiting for the element.

Another options is to be able to wait for multiple items on the page and ensure that always at least one is distinguishing.

What do you think? I would possibly do both since they address slightly different cases.

witzig: In hindsight, I'm not sure whether using elements for the catch all `waitUntilVisible` was the right decision, at least for the subscription process. Perhaps it would make more sense to [wait for](https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/lib/until.html) unique text, e.g. "Subscription Confirmed". This way we'd also get an essential text assertion for free and the links/buttons are tested no later than we're trying to click them.

> What do you think? I would possibly do both since they address slightly different cases.

Sounds good. `waitUntilVisibleAfterRefresh ` would solve this problem nicely. Waiting for multiple items makes sense too, yet (for this particular issue) there aren't many readily identifiable elements in the HTML generated by MJML.

bures:
@witzig - There a few more tests to that I want to do. Also I would like to fix the issue with the "auto" parameter.
However, this is what I believe could be more or less the way tests could look like.

I added the two things we discussed above. Fixed the reporting a bit. 

I introduced functions that save the current page (as HTML, url, and PNG). This can be invoked via API. It also happens automatically when a test fails - it gets stored in last-failed-e2e-test.xxx. This is quite useful when creating and debugging the tests.

I also extended the API of "useCase". It can be now called with .only, .skip - the same was as "it"/"test" can be marked with .skip/.only.

witzig:
Tomas, this is just fantastic. I'm greatly amused watching Firefox executing all these tests. :-)

> There a few more tests to that I want to do.

You've already dedicated quite some time towards the e2e tests, so I'd be happy to help with the remaining use cases you described. It's mostly assiduity work anyway and I think the foundation is laid and quite complete.

Regarding the List Unsubscribe Header, it's only been implemented recently with 745bfb0. Also here's a more detailed explanation [how it works with gmail](http://www.zettasphere.com/gmail-email-unsubscribe-header/). Given the example with the emails from Google Analytics, it looks like the auto parameter isn't too important after all ...

bures:
Perfect. Thanks. In that case, I'll only  implement the "auto" parameter in the afternoon and kindly ask you for help with the remaining tests.

How shall we do it. Merge this PR first once the "auto" parameter is in?
Or move it to a branch and finish it there? Or I can give you developer access to my fork (bures/mailtrain). As you like. Thanks.

witzig:
We both tested the remaining use cases already manually. So, yes, let's merge this once the "auto" parameter is in.

bures:
I'm a bit puzzled now. The List-Unsubscribe header has been introduced by 745bfb0, but it has later been deleted in 179e0d5ab39bfede012849e9a1a49f4e27802a39. 

@andris9 - it seems that the deletion was by you. Was it intentional?

bures:
I've added handling of the "auto" parameter to the unsubscription. However, the List-Unsubscribe header is currently not included in the emails (see my comment above).

andris9:
@bures I think the List-Unsubscribe was removed because of custom unsubscribe links. Some users had their own custom unsubscribe pages and updated lists via API so they did not want the link to be there. Now I think that removing the link was a mistake, instead it should have been configurable option or something like that. I'm in favor putting the link back to the headers as default.

witzig:
I'm a bit puzzled too, as the List-Unsubscribe Header is still present. I reckon it's [generated by Nodemailer](https://community.nodemailer.com/2-0-0-beta/list-headers/) based on [this](https://github.com/Mailtrain-org/mailtrain/blob/684ba69d467e7167a09a95497e2ff8b386c48e6a/services/sender.js#L420-L422). @andris9, might this have been the reason for the removal in 179e0d5?

witzig:
@bures, could you add your git email to github, so your commits get properly linked and attributed? Thanks.

bures:
@witzig, good point. Can you please check now. I added the address to github. Thanks.

witzig:
@bures, your commits are now properly linked. Thanks.

witzig:
I pushed the remaining tests with 19a563b. Any other scenarios we should check?

Thoughts about integration with https://travis-ci.org?

bures:
@witzig - Thanks a lot for the remaining tests. I've fixed a few eslint problems and I'm commiting it now. 
As for the switchToList, I updated it such that the page object is constructed for every use-case. I believe this is cleaner as they are now self-contained.

Travis may be be good. Also it would make sense to automate the execution of eslint. I admit, I forgot about it when doing the last PR.

witzig:
@bures, no worries regarding eslint. I'm guilty myself about forgetting it, Andris can tell. :-)

Also, I should have told you that I already fixed the errors. I assumed merging master would be your first move before continuing work on the e2e stuff. Sorry about that. Fixing the conflicts in 731226d surely wasn't fun. :/

> Also it would make sense to automate the execution of eslint.

#242 would take care of that. And perhaps a pre-commit hook could be an option, too.

bures:
>  I assumed merging master would be your first move before continuing work on the e2e stuff.
Haha. That would be of course very natural thing to do if I had not forgotten about your commit (which I of course saw the day before). :-D

andris9:
@witzig oh, yes, it seems that I moved the List header generation from one place to another and did not just remove it. Already forget about it. I think I re-implemented [List- header handling](https://nodemailer.com/message/list-headers/) for Nodemailer and wanted to test it with Mailtrain.

Issue #75, opened 2016-07-13
Export List


luisacamara:
Hello, how can I export the list from mailtrain please?

andris9:
Right now there is no export function in Mailtrain. You need to export subscribers from MySQL database directly. All required fields can be found from table `subscription__{ID}` where `{ID}` is the numeric ID of the list. If you have found out the numeric ID, then you can query all subscribers from the database like this (assuming that list ID is 2):

    SELECT last_name, first_name, email FROM subscription__2 WHERE status=1 ORDER BY last_name, first_name, email

luisacamara:
many thanks for your help.



Kind regards/ Saludos,


[image: Logo XML Travelgate] <http://www.xmltravelgate.com/>

Luisa Camara

Optimization Executive - XML Travelgate

Tel: + 34 871 968 181 | Ext: 104

lcamara@xmltravelgate.com | www.xmltravelgate.com

<https://www.facebook.com/XmlTravelgate>
<https://twitter.com/XMLTravelgate>
<http://www.linkedin.com/company/xml-travelgate>
<https://plus.google.com/+Xmltravelgate/posts>

programster:
@andris9 this looks like a really useful feature to have (I could do with having it right now). Would you like me to have a whack at creating a pull request for it at some point or are you already working on it or deliberately don't want this feature in the app?

andris9:
If you could work on a PR it would great. I don't know that anyone would be already working on this and I'm not very active currently with Mailtrain myself, mostly just review pull requests etc.

programster:
@andris9 sounds good. It would probably come at the weekend depending when I get the time.

bures:
@programster  See the new "Reports" functionality. It allows generating a "CSV" report, which can be used to export a list. There is a sample Report template for that purpose there.

Any feedback welcome. Thanks.

programster:
Hey @bures I'm being really thick and can't find the reports functionality although I can clearly see there is a reports.js view and stuff. What would the url be like? e.g. if I go to my-site.org/reports I get a not found message. Perhaps it's hidden or an API request I need to make?

programster:
Just realized that I misunderstood your message. Your pointing out how I should start on the feature, not telling me that you have gone and done it.  :)

bures:
[...]

PR #255, opened 2017-06-15
Docker updates

tiangolo:
* Update `Dockerfile` to be based on an official Node JS image as this is a Node JS project.
* Add a working `docker-compose.yml` file (the one in the examples doesn't work as it has some syntax issues).
* Modify the Docker version to use a default `docker.toml` file with sensible configurations, compatible with `docker-compose.yml`, allowing a faster and more straightforward deployment.
* Add a `docker-compose.override.yml` file with overrides to build the image, map ports, and `restart: always`.
* Update the `README.md` to include instructions taking into account the new Docker changes.


witzig:
@tiangolo, thanks for your work.

> Update Dockerfile to be based on an official Node JS image as this is a Node JS project.

I'm not sure why centos was initially chosen, but I guess the official Node JS image does lack ImageMagick, which is used by Mosaico and GrapeJS (e.g. jquery-file-upload-middleware).

> Modify the Docker version to use a default docker.toml file with sensible configurations, compatible with docker-compose.yml, allowing a faster and more straightforward deployment.

`docker.toml` is basically a copy of `default.toml` with the following changes:

```
[mysql]
host="mysql"

[redis]
enabled=true
host="redis"
```

While it does 'allow for a faster and more straightforward deployment', I don't think `docker.toml` should be committed (as is). Same for `docker-compose.override.yml`. The reasoning being:

> You might want to modify the `docker-compose.yml` or `docker-compose.override.yml` file, modify port mappings, change volume paths, etc.

If a file is intended to be changed by the user, it probably shouldn't be committed (IMHO).

Pinging @larrabee and @programster, the docker guys. What do you think?

larrabee:
Hello. Hmm, why you prefer official node.js docker image? Official node:8.1 is too heavy (265 MB) vs  130 MB (centos + nodejs). node:8.1 based on Debian jessie, i dont know why we should change centos to debian.

programster:
To me, the most important factors are:
* does the environment work (image magick, supported versions, deprecations etc).
* the environment needs to reflect what the main _maintainers_ are most comfortable with supporting and are actually dev/testing in. If the maintainers are developing/testing in CentOS and don't want to switch to Debian, then the environment shouldn't change (I do not consider myself an official maintainer as I have not reached that level of commitment yet).

I personally would welcome a switch from CentOS to Debian which it appears is what the official node image is based upon, as I have a lot more experience with debian, but I am not one of the main maintainers. I would be happy to assist with getting packages such as ImageMagick installed in the Dockerfile if necessary.

### Size
I do not think size matters (I tell all the ladies that). If you build the image, size really doesn't matter, and if you're pulling the image, then usually these are in datacenter environments that have the bandwidth and you only have to pull like once every blue moon. 


### Expected Changes/Configurations
> If a file is intended to be changed by the user, it probably shouldn't be committed (IMHO).

A general good practice for files that are expected to be filled-out/changed by users is to just add the `.temp` (or other appropriate suffix) to represent a template file. This file should have lots of comments telling the user about what they shouldd put in. E.g a `settings.php.tmpl` file that would have the following contents:

```
# Put in the hostname of your database here...
define('DB_HOSTNAME, '');
```
The readme fille then tells the user to remove the extension for the project to work. The file without the extension should be added to .gitignore so that they are never accidentally committed.

witzig:
Thank you both. To me, you guys actually are the maintainers of all things Docker, also since Andris was long reluctant to even merge a Dockerfile. 

I scanned the network graph for Docker and Heroku forks and found the following:

1. Docker: https://github.com/Mailtrain-org/mailtrain/compare/master...djalal:master
2. Heroku: https://github.com/Mailtrain-org/mailtrain/compare/master...jaredpalmer:master
3. Heroku: https://github.com/Mailtrain-org/mailtrain/compare/master...potherca-contrib:feature/deploy-to-heroku

Fork 1 and 2 have in common that they commit config files which shouldn't be committed (as does this PR), fork 3 uses [custom-environment-variables.json](https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables) instead, although the 'custom' prefix implies that this probably shouldn't be committed either.

Perhaps a solution could be to convert `default.toml` to `default.js`, then this would be possible:

```
mysql: {
  host: process.env.MAILTRAIN_MYSQL_HOST || 'localhost'
}
```

In `docker-compose.yml` we'd set these environment variables which can be overwritten with `docker-compose.override.yml`. Using an external `production.toml` (which makes sense to me) would require a custom `docker-compose.override.yml` for mounting the file.

In short: By converting `default.toml` to `default.js` and committing `docker-compose.yml` we could implement what this PR brings to the table: "allowing a faster and more straightforward deployment".

The README should advise to use a custom `docker-compose.override.yml` for modifications, and also explain how to mount an external/custom config file.

Yet before someone starts to work on this ... @andris9, @bures, thoughts/objections?

andris9:
Just to point out that if you change NODE_ENV from production to something else you might loose the [optimisations by Express.js](https://expressjs.com/en/advanced/best-practice-performance.html#in-environment).

tiangolo:
Thank you everyone for your quick feedback!

@witzig :

* ImageMagick is already installed in the official Node JS image (I checked that before the PR). That image also has building dependencies included, so anyone that needs to support binary extensions could easily do that too.
* About the image size and Debian: basing an image in Debian [is a "best practices" recommendation by Docker](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#from) . Because of that, most of the official images are based on Debian. So, if a user has to download [`mysql`](https://github.com/docker-library/mysql/blob/6b1dc54320b03b83a89068f49cc796fea0ff6bb4/5.7/Dockerfile#L1) and [`redis`](https://github.com/docker-library/redis/blob/fa0e115934f352cbfa8d0d471b43a18e865e28b3/3.2/Dockerfile#L1) they are already downloading and having all the Debian layers on their system. And as you know, Docker images are made of shared layers, so, when pulling the official NodeJS image, users are normally only pulling the layers that go on top of Debian (the NodeJS installation) as they already have the base layers for MySQL and Redis. 
  * Thinking of new developers that come to contribute to the project, because of the same (official Docker images based on Debian), it's probably more common to find Docker developers that are familiar with Debian than with CentOS. And also, I imagine that NodeJS Docker developers would be more familiar with official NodeJS images.
  * I agree with @programster that size is not that important, and I've also been convincing my girlfriend of that... still, if you have strong concerns about image size (despite the fact that users are already downloading those layers for MySQL and Redis), I could suggest using an Alpine variant of the official NodeJS image. That way you get to be based on a NodeJS official version but a very small one (64 MB). Nevertheless, if you don't have a specific reason why to worry that much about image size, I wouldn't recommend that to you. Alpine is a more container-specific distro, with some different commands, etc. So, if you have not been a very Docker-centric project with some specific needs to have a very specific setup (requiring to use Alpine) I would suggest you keep the most simple and generalizable option, I would say a standard official NodeJS image.

* About committing config files, if you want, I can change the `docker-compose.override.yml` to a `.tmpl` file, so that way people can have a working example to start from. I think that would be better than just adding it in the README to avoid things like, the current example in the README doesn't work because it has some indentation syntax errors, and debugging that was not quite easy as the content seems "correct". But being inside of a README would be very easy to accidentally break the format again.
* About committing the `docker.toml` and to handle the optimizations pointed out by @andris9 , let me suggest the following:
  * Commit a `production.toml.tmpl` with the Docker Compose compatible configurations.
  * Add a Docker `entrypoint.sh` that checks if there exists a `production.toml`, and if not, copy the `production.toml.tmpl` to it. That way, when someone modifies and sets up a customized `production.toml` it will be used. But if they didn't, they still can have a working environment. Also, doing that, the `NODE_ENV` can be set to `production` again.

* @witzig I would suggest not mounting a file from the host, as it would make it work only for a single node, you couldn't run in a distributed Docker system (Rancher, Kubernetes) so easily, but that's just a suggestion.

---

What do you guys think of doing the following?:

* Keep the official standard NodeJS image
* Set `NODE_ENV` to `production`
* Add a `production.toml.tmpl` with the configs that are in `docker.toml`
* Add an `entrypoint.sh` that copies `production.toml.tmpl` to `production.toml` when there is not one already
* Change `docker-compose.override.yml` to `docker-compose.override.yml.tmpl` and add instructions in the `README` on how to change and use it.

witzig:
Sounds good to me. The only thing I would still suggest is to keep `production.toml.tmpl` minimal, like it's advised in [default.toml](https://github.com/Mailtrain-org/mailtrain/blob/master/config/default.toml#L7) and handled in [install.sh](https://github.com/Mailtrain-org/mailtrain/blob/master/setup/install.sh#L78-L90). And perhaps name it differently to avoid confusion, maybe add a `docker` infix.

It would also be great if we could set up (and enable?) the [reports feature](https://github.com/Mailtrain-org/mailtrain/blob/master/config/default.toml#L159-L172). It writes to `/app/protected/reports`. What do you think?

bures:
Sounds good to me too. The reports feature would be nice to have there so that we ship Mailtrain with all its capabilities.

programster:
Sounds good to me.


> ...I could suggest using an Alpine variant of the official NodeJS image...

Oh god please no. 

As @bures mentioned, I would love to get reports in the out of the box web interface (hopefully this includes the abilty to export csv lists). [I never got round to getting onto this](https://github.com/Mailtrain-org/mailtrain/issues/75), and in the end just [made my own PHP script that hooked up to the mailtrain database](https://github.com/programster/mailtrain-tools) to grab the data and export to a CSV for me. I know that is a less-than-stellar solution but I needed something quick that I could easily understand and do in 5 minutes when I was in a bind.

tiangolo:
Thank you all for your valuable feedback!

Here's what I did:

* Keep the official standard NodeJS image
* Set `NODE_ENV` to `production`
* Add a `docker-production.toml.tmpl` with the minimal needed configs to be used with `docker-compose` on top of `default.toml`. And enable reports in it.
* Add a `docker-entrypoint.sh` that copies `docker-production.toml.tmpl` to `production.toml` when there is not one already, but uses the one existing when there's one.
* Make the `docker-entrypoint.sh` copy the `docker-production.toml.tmpl` to `production.toml` inside of the reports configs, as it uses its own configs and the reports wouldn't work without it. 
* Make the `docker-entrypoint.sh` copy the `production.toml` file in `config` (when it exists) to the reports configs.
Change `docker-compose.override.yml` to `docker-compose.override.yml.tmpl` and add instructions in the README on how to change and use it.
* Add a volume for the reports' data.

---

I confirm I tested it and it worked. 

Reports worked like a charm (nice feature! I didn't know about that). 

And I confirm that the entrypoint is doing what it should:
* Copy the template config to the main config directory as `production.toml` only when needed. 
* Respect if the user provided a different `production.toml` in the main config directory.
* Use the provided `production.toml` from the main config directory for the reports config (instead of the template) when there's one in the main config directory.
* Use any `production.toml` that is in the reports config directory when there's one there (instead of the `production.toml` in the main config directory or the template).

witzig:
Thanks! Looks good to me. Let's give @bures, @programster and @larrabee some time to review too.

larrabee:
Hello. Looking good for me too.

tiangolo:
Awesome! Thanks! :tada:

PR #269, opened 2017-06-27
Check for locally requeued messages in postfix

losacno:
When using Postfix + Amavis received mail first goes to Amavisd and, if not spam, is queued to Postfix Smtp service so queue id for bounce check is not the first one but the second as you can see:

    Jun 27 14:30:11 server-mx postfix/smtpd[8270]: connect from localhost[127.0.0.1]
    Jun 27 14:30:11 server-mx postfix/smtpd[8270]: NOQUEUE: filter: RCPT from localhost[127.0.0.1]: <mail@sender.tld>: Sender address triggers FILTER amavis:[127.0.0.1]:10026; from=<mail@sender.tld> to=<mail@receiver.tld> proto=ESMTP helo=<[127.0.0.1]>
    Jun 27 14:30:11 server-mx postfix/smtpd[8270]: B2BBE8414B5: client=localhost[127.0.0.1], sasl_method=PLAIN, sasl_username=mail@sender.tld
    Jun 27 14:30:11 server-mx postfix/cleanup[8313]: B2BBE8414B5: message-id=<mail-id@sender.tld>
    Jun 27 14:30:11 server-mx postfix/qmgr[1694]: B2BBE8414B5: from=<mail@sender.tld>, size=123137, nrcpt=1 (queue active)
    Jun 27 14:30:12 server-mx postfix/smtpd[8318]: connect from localhost[127.0.0.1]
    Jun 27 14:30:12 server-mx postfix/smtpd[8318]: C642C8414CE: client=localhost[127.0.0.1]
    Jun 27 14:30:12 server-mx postfix/cleanup[8313]: C642C8414CE: message-id=<mail-id@sender.tld>
    Jun 27 14:30:12 server-mx postfix/smtpd[8318]: disconnect from localhost[127.0.0.1]
    Jun 27 14:30:12 server-mx postfix/qmgr[1694]: C642C8414CE: from=<mail@sender.tld>, size=124068, nrcpt=1 (queue active)
    Jun 27 14:30:12 server-mx amavis[5085]: (05085-15) Passed CLEAN {RelayedOutbound}, ORIGINATING LOCAL [127.0.0.1]:45421 <mail@sender.tld> -> <mail@receiver.tld>, Queue-ID: B2BBE8414B5, Message-ID: <mail-id@sender.tld>, mail_id: KFebp5elSvWC, Hits: -0.997, size: 123137, queued_as: C642C8414CE, dkim_new=dkim:sender.tld, 1048 ms
    Jun 27 14:30:12 server-mx postfix/smtp[8314]: B2BBE8414B5: to=<mail@receiver.tld>, relay=127.0.0.1[127.0.0.1]:10026, delay=1.2, delays=0.11/0.01/0/1.1, dsn=2.0.0, status=sent (250 2.0.0 from MTA(smtp:[127.0.0.1]:10027): 250 2.0.0 Ok: queued as C642C8414CE)
    Jun 27 14:30:12 server-mx postfix/qmgr[1694]: B2BBE8414B5: removed
    Jun 27 14:30:15 server-mx postfix/smtp[8320]: C642C8414CE: to=<mail@receiver.tld>, relay=remote-mx.tld[2.2.2.2]:25, delay=2.5, delays=0.07/0.01/2.1/0.26, dsn=5.7.1, status=bounced (host remote-mx.tld[2.2.2.2] said: 550 5.7.1 <mail@receiver.tld>: Recipient address rejected: User unknown. (in reply to RCPT TO command))
    Jun 27 14:30:15 server-mx postfix/cleanup[8313]: 6DE8F841563: message-id=<new-mail-id@server.sender.tld>
    Jun 27 14:30:15 server-mx postfix/bounce[8324]:C642C8414CE: sender non-delivery notification: 6DE8F841563

This PR try to solve this problem saving locally requeued messages in an array and checking bounces against it

### DIFF
--- a/lib/models/campaigns.js
+++ b/lib/models/campaigns.js
@@ -1086,6 +1086,24 @@ module.exports.updateMessage = (message, status, updateSubscription, callback) =
     });
 };
 
+module.exports.updateMessageResponse = (message, response, response_id, callback) => {
+    db.getConnection((err, connection) => {
+        if (err) {
+            return callback(err);
+        }
+
+        let query = 'UPDATE `campaign__' + message.campaign + '` SET `response`=?, `response_id`=? WHERE id=? LIMIT 1';
+        connection.query(query, [response, response_id, message.id], err => {
+            connection.release();
+            if (err) {
+                return callback(err);
+            }
+            return callback(null, true);
+        });
+
+    });
+};
+
 function createCampaignTables(id, callback) {
     let query = 'CREATE TABLE `campaign__' + id + '` LIKE campaign';
     db.getConnection((err, connection) => {
--- a/services/postfix-bounce-server.js
+++ b/services/postfix-bounce-server.js
@@ -41,18 +41,16 @@ let server = net.createServer(socket => {
                 // Losacno: Check for local requeue
                 let status = match[2];
                 log.verbose('POSTFIXBOUNCE', 'Checking message %s for local requeue (status: %s)', queueId, status);
-                if ( status == 'sent' ) {
+                if ( status === 'sent' ) {
                     let queued = / relay=127\.0\.0\.1/.test(line) && line.match(/ queued as (\w+)\)/);
                     if ( queued ) {
                         log.verbose('POSTFIXBOUNCE', 'Marked message %s as locally requeued as %s', queueId, queued[1]);
                         queueIds[queued[1]] = queueId;
                     }
                     return checkNextLine();
-                } else {
-                    if ( queueId in queueIds ) {
-                        log.verbose('POSTFIXBOUNCE', 'Message %s was requeued from %s', queueId, queueIds[queueId]);
-                        queueId = queueIds[queueId];
-                    }
+                } else if ( queueId in queueIds ) {
+                    log.verbose('POSTFIXBOUNCE', 'Message %s was requeued from %s', queueId, queueIds[queueId]);
+                    queueId = queueIds[queueId];
                 }
 
                 campaigns.findMailByResponse(queueId, (err, message) => {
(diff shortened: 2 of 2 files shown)


witzig:
Thanks @losacno.

From a quick glance, it seems there's a memory leak regarding `queueIds` and `seenIds`. This might have been negligible so far for `seenIds` as it only stored bounces, but with the proposed change this looks like a problem to me.

 I'll leave this for @andris9 to review.

witzig:
@losacno, with your PR, processing the postfix logs of 1 million successfully sent emails will raise memory usage quite a bit. Testing with:

```javascript
let seenIds = new Set();
let queueIds = {};

console.log(process.memoryUsage().rss / 1048576 + ' MB');

for (let i = 0; i < 1000000; i++) {
    let id = '#' + (1000000000 + i);
    seenIds.add(id);
    queueIds[id] = id;
}

console.log(process.memoryUsage().rss / 1048576 + ' MB');
```

Logs:

```
27.16796875 MB
361.02734375 MB
```

I think the `seenIds` and `queueIds` will have to expire somehow. Perhaps [memory-cache](https://github.com/ptarjan/node-cache) could be an easy fix for this problem, yet there might be a better solution around.

mfechner:
I tested it and at least it fixes the problem I have described in #277

mfechner:
I thought a little bit about that problem.
And the queue ID is stored in the database which is in this case the internal queue ID.
What is, if we would update the internal queue ID and overwrite it with the real (to external mailserver) queue ID in the database.
If we would have the real queue ID (to the external mailserver) in the database, the normal bounce handler should be able to find the correct entry and can deactivate the email correctly.

What do you think about that?

witzig:
That sounds like a pragmatic solution to me.

Would you be willing to create a PR for this, if there're no objections after a few days?

losacno:
Thanks for yours notes, i changed a bit the commit to save new queue_id in database and to free up memory after processing ids

andris9:
@losacno otherwise looks good, I'm not use about the hardcoded `127\.0\.0\.1` regex though (ie. the relay could be just as well `::1`). Maybe ignore the relay address entirely: `/ relay=/.test(line)` ?

losacno:
you're right, removed.

PR #275, opened 2017-07-04
Added support for an `update` end-point in the API.

Roost:
This allows a change to data for an existing subscription, without needing to know the subscription state of the email account- useful when calling the API from a server who has no knowledge of the state of the existing subscription. This assumes the email account is already known.

### DIFF
--- a/routes/api.js
+++ b/routes/api.js
@@ -315,6 +315,126 @@ router.post('/delete/:listId', (req, res) => {
     });
 });
 
+router.post('/update/:listId', (req, res) => {
+    let input = {};
+    Object.keys(req.body).forEach(key => {
+        input[(key || '').toString().trim().toUpperCase()] = (req.body[key] || '').toString().trim();
+    });
+    lists.getByCid(req.params.listId, (err, list) => {
+        if (err) {
+            log.error('API', err);
+            res.status(500);
+            return res.json({
+                error: err.message || err,
+                data: []
+            });
+        }
+        if (!list) {
+            res.status(404);
+            return res.json({
+                error: 'Selected listId not found',
+                data: []
+            });
+        }
+        if (!input.EMAIL) {
+            res.status(400);
+            return res.json({
+                error: 'Missing EMAIL',
+                data: []
+            });
+        }
+
+        tools.validateEmail(input.EMAIL, false, err => {
+
+            if (err) {
+                log.error('API', err);
+                res.status(400);
+                return res.json({
+                    error: err.message || err,
+                    data: {
+                        success: false
+                    }
+                });
+            }
+
+            subscriptions.getByEmail(list.id, input.EMAIL, (err, subscription) => {
+
+                if (err) {
+                    log.error('API', err);
+                    res.status(400);
+                    return res.json({
+                        error: err.message || err,
+                        data: {
+                            success: false
+                        }
+                    });
+                }
+
+                let updateObject = {
+                    email: input.EMAIL
+                };
+
+                if (input.FIRST_NAME) {
+                    updateObject.first_name = (input.FIRST_NAME || '').toString().trim();
+                }
+
+                if (input.LAST_NAME) {
+                    updateObject.last_name = (input.LAST_NAME || '').toString().trim();
+                }
+
+                if (input.TIMEZONE) {
+                    updateObject.tz = (input.TIMEZONE || '').toString().trim();
+                }
+
+                fields.list(list.id, (err, fieldList) => {
+
+                    if (err && !fieldList) {
+                        fieldList = [];
+                    }
+
+                    fieldList.forEach(field => {
+
+                        if (input.hasOwnProperty(field.key) && field.column) {
+                            updateObject[field.column] = input[field.key];
+                        } else if (field.options) {
+                            for (let i = 0, len = field.options.length; i < len; i++) {
+                                if (input.hasOwnProperty(field.options[i].key) && field.options[i].column) {
+                                    let value = input[field.options[i].key];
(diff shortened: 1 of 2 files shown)


bures:
Nice. Thanks.

Could we also have a documentation to the endpoint? At least in views/users/api.hbs. Thanks.

witzig:
Can you elaborate on the use case of this a bit more? Is the shortcoming of the endpoint `/api/subscribe/:listId`, which "either inserts a new subscription or updates existing", the state when a confirmation is pending?

From a quick glance, the new endpoint seems to share most of its code with `/api/subscribe/:listId` (and is also missing a check [here](https://github.com/Roost/mailtrain/blob/920e41d26eb2a0efa71ec2621e7ea7119f06d8a7/routes/api.js#L372) whether the subscription actually exists, e.g. like [this](https://github.com/Mailtrain-org/mailtrain/blob/87f7a050ab5882f71b4dd535777b5fed19286bfe/routes/api.js#L221-L227)).

Thus I'm wondering if an endpoint for retrieving the state of a subscription would be more beneficial.

Roost:
After further experimentation, I have realised that the `subscribe/:listId` endpoint won't take a user from an `unsubscribed` status to a `subscribed` status- which feels wrong. My initial assumption was that I didn't want my application server to have any knowledge about the subscriptions status, leaving all that logic to the Mailtrain server. Hence, I had designed my requests to the Mailtrain server to exclude the subscription status, which meant I needed a new end point to update custom fields linked to automated triggers. 

Now, in my opinion, we have 2 issues;

1. If I call `subscribe/:listId`, I would expect it to explicitly mark the subscription status as `subscribed`. Imagine a scenario where a user has previously unsubscribed and then tries to resubscribe via another means. If you believe the current behaviour is correct, then I would at least expect the response to notify me that the subscription state was not changed as expected.
2. It seems the the status of a subscription should not be linked to the API endpoint, but rather passed as an optional parameter into an API (maybe `update/:listId`) so we can make changes to a subscription without effecting it's status. For example:

```
{
"EMAIL": "some@coool.com",
"ACTION": "subscribe/unsubscribe/update/delete/whatever",
"MERGE_CUSTOM_FIELD" : "whatever"
}
```

Thoughts @witzig?

witzig:
>  If I call subscribe/:listId, I would expect it to explicitly mark the subscription status as subscribed. 

Currently, you can resubscribe an email with the endpoint `subscribe/:listId` while setting `FORCE_SUBSCRIBE` or `REQUIRE_CONFIRMATION` to `yes` (although when using the latter it looks like we're losing existing custom field data, which is another issue).

> If you believe the current behaviour is correct, then I would at least expect the response to notify me that the subscription state was not changed as expected.

I reckon the problem is to know the subscription state (i.e. `unsubscribed`) in the first place. This might be complicated by using the same endpoint for `create` and `update`, which is kind of set in stone however, i.e. backwards compatibility.

> It seems the the status of a subscription should not be linked to the API endpoint, but rather passed as an optional parameter into an API (maybe update/:listId) so we can make changes to a subscription without effecting it's status.

I'm not sure if we have a misunderstanding here, but you can already make changes to a subscription without affecting its status. If you call `subscribe/:listId` without `FORCE_SUBSCRIBE` or `REQUIRE_CONFIRMATION`, the subscription status is unaffected.

Some examples, POSTing to `subscribe/:listId`:

1. Add subscription: `{ EMAIL: 'foo@example.com', F1: 'custom field one', F2: 'two'  }`. If the email doesn't already exist, a new subscription is added with status `subscribed`. If it exists, this API call will only update `F1` and `F2`.

2. Update subscription: `{ EMAIL: 'foo@example.com', F1: 'new value'  }`. Only `F1` is changed.

3. The subscriber unsubscribes somehow.

4. To resubscribe via API, you need to post `{ EMAIL: 'foo@example.com', FORCE_SUBSCRIBE: 'yes' }`. 

To me, it seems that some API actions simply require knowledge about the subscription, e.g. when to offer a resubscribe button on your custom account page. Thus the suggestion for an additional query endpoint.

> My initial assumption was that I didn't want my application server to have any knowledge about the subscriptions status, leaving all that logic to the Mailtrain server. Hence, I had designed my requests to the Mailtrain server to exclude the subscription status, which meant I needed a new end point to update custom fields linked to automated triggers.

I still don't really understand why you need a new endpoint for this.

Roost:
Thanks for the update, and I think I need to further clarify that I am indeed able to do the required actions via the current API implementation, but I am suggesting that the current structure is not ideal and could be improved.

Following the logic of using the `FORCE_SUBSCRIBE` parameter in the `subscribe/:listId` API, it stands to reason that should I want to unsubscribe a subscription via the API, I should pass `FORCE_UNSUBSCRIBE` as a parameter to the `unsubscribe/:listId`? 

What I am suggesting is that having a single entry point with an appropriate switch based on the supplied `ACTION` parameter could make the entire API simpler to use and understand:

A suggested example;

Endpoint:
 `http://{{baseUrl}}/api/:listId` 

Parameters
```
{
    EMAIL: 'foo@example.com', 
    ACTION: 'subscribe`
    F1: 'new value' 
}
```
would have the same behavior as the current implementation using `FORCE_SUBSCRIBE`, while calling the suggested endpoint with `ACTION: 'update'`, will do the update instead. Similarly for `DELETE` etc.

It just strikes me as odd that an API named `subscribe` needs an additional parameter to actually do it's intended task.

andris9:
The current API design is intentional. It is like this to discourage programmatic resubsribing users that have unsubscribed from the list. You can still keep updating user data but to actually resubscribe the user must do it himself through subscription form and email verification or you need to use the FORCE option.

PR #282, opened 2017-07-21
Fix docker install instructions

FWirtz:
I just tried to install Mailtrain using Docker.
I suppose this line from the install instructions contains an error:

> Copy the file docker-compose.override.yml.tmpl to docker-compose.override.yml.tmpl and modify it if you need to.

I suppose the destination filename should not contain the `.tmpl` ending. Is that correct? If so, I changed it accordingly.

`Change extension name of copied docker-compose-override file to not be the same as source`


witzig:
Thanks!

Issue #292, opened 2017-08-14
Documenting DKIM setup


killua-eu:
Hi,

While rewriting maitrain's installation script into an an ansible playbook, I got into setting up DKIM properly. The current documentation isn't really helpful so I added some to the readme of https://github.com/Vaizard/mage-mailtrain. Please feel free to cherry pick what you think is relevant, or let me know and i'll just create a PR on mailtrain's readme myself.

thanks for all the work, P.

andris9:
Hey, that's great. Could you make a PR out of it?

killua-eu:
Done.

PR #293, opened 2017-08-14
Improving DKIM/SPF/DMARC/PTR documentation

killua-eu:
This should resolve https://github.com/Mailtrain-org/mailtrain/issues/292


andris9:
Thanks!

Issue #141, opened 2017-02-25
Install error - bash: line 15: apt-add-repository: command not found


rnarayan:
When I ran the command (on a brand new Linux 16.04 LTS server at Linode)
`curl https://raw.githubusercontent.com/andris9/mailtrain/master/setup/install.sh | sudo bash`

I got the following error
`bash: line 15: apt-add-repository: command not found`

I found the solution at this blog - https://pricklytech.wordpress.com/2014/05/16/ubuntu-server-14-4-trusty-add-apt-repository-command-not-found/

I ran the following command and the problem was solved.
`sudo apt-get install software-properties-common`

Would it be possible to add a check if apt-add-repository exists and if not, install it before running the next command in bash script.

forgeOneHQ:
@rnarayan if you still use it please try v2 which is already stable.

PR #314, opened 2017-09-18
Update install.sh

wotomg:
Some fixes. Anyway, scripts does not work.


witzig:
Hello @wotomg, thanks for this PR. Two questions:

1. Wouldn't it be possible to set an empty MariaDB root password?

2. You didn't run into the error reported in #141, `apt-add-repository: command not found`?

Issue #325, opened 2017-09-27
Can list import be made to preserve unsubscribed status?


deisner:
We're considering Mailtrain for a newsletter that will have periodic additions to the recipient list via CSV import. I've installed Mailtrain, and it looks like the CSV import clobbers the unsubscribed status field. Is there a way to prevent this, or workaround it somehow?

My concern is that a user who unsubscribes herself may accidentally be added back at a later date.

And by the way, thanks for Mailtrain -- it really looks promising.

witzig:
Yeah, that's a reasonable concern. We can add an additional option to the import screen.

Categorize the imported subscribers as:

- Subscribed – Regular subscriber addresses
- Subscribed (Force) – Regular subscriber addresses, resubscribe users that have unsubscribed
- Unsubscribed – Suppressed emails that will be unsubscribed from your list

deisner:
That was fast. I pulled your commit and it seems to work as expected -- nothing "funny". :^) Danke.

witzig:
Ha! Gern geschehen. :)

PR #326, opened 2017-09-27
Add option to preserve unsubscribed status on list import

witzig:
References issue #325

Tomas, can you review this please? Using `0` as the subscription status leaves the status of an existing subscription unchanged or defaults to `SUBSCRIBED` for new users. See [here](https://github.com/Mailtrain-org/mailtrain/blob/4f5c132db4bca35e0fb310ce1c03da73cd9e9170/lib/models/subscriptions.js#L165): 

```
meta.status = meta.status || (existing ? existing.status : Status.SUBSCRIBED);
```

It feels a bit like cheating to me. :)

- - -

screenshot

### DIFF
--- a/lib/models/subscriptions.js
+++ b/lib/models/subscriptions.js
@@ -625,7 +625,7 @@ module.exports.delete = (listId, cid, callback) => {
 
 module.exports.createImport = (listId, type, path, size, delimiter, emailcheck, mapping, callback) => {
     listId = Number(listId) || 0;
-    type = Number(type) || 1;
+    type = Number(type) || 0;
 
     if (listId < 1) {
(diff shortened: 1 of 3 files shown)


bures:
Hello Roger. 
I believe it's OK for now. I'm likely to refactor this part a bit in the near future for the permissions/React patch. I'll see whether I can have something like a flag "keepExistingStatus".

For now, I would keep it as you did it. It's a solid and working solution. If you could just put there a comment that this may need to be revisited.

Thanks.
T.

witzig:
Thanks Tomas, I added the comment for you. I also realised the API endpoint `subscribe` makes use of the same feature as to not resubscribe a user unless using FORCE_SUBSCRIBE. :)

Issue #340, opened 2017-10-10
[LDAP auth] No differentiation between "bind" user and "search" user


MelwinKfr:
I'm currently facing to an LDAP security strategy which doesn't allow every user to search, it requires to be an admin user. So I currently cannot connect any user using this LDAP.

This [pull request](https://github.com/tphuocthai/passport-ldapjs/pull/4) allows more options for LDAP config and was already merged into `tphuocthai/passport-ldapjs/master`.

Would it be possible to integrate those new options into Mailtrain ?

PR #341, opened 2017-10-10
add: LDAP options `bindUser` and `bindPassword`

MelwinKfr:
To be able to use a different user for bind requests,
which can be required by some LDAP security strategy.

Linked to issue #340


witzig:
Thanks for this PR, @MelwinKfr.

Do you think we should add some documentation to the config file? [See here](https://github.com/Mailtrain-org/mailtrain/blob/master/config/default.toml#L114-L123).

MelwinKfr:
Absolutely, here is the change.

witzig:
Thanks!

Issue #247, opened 2017-06-01
Feature Suggestion - Duplicate Campaign/Template


ubergeekady:
I think the ability to copy a campaign / duplicate it, is an essential feature in most email campaign management applications. I have used it multiple times in the past with other tools. Say for example, i want to send the same campaign to multiple lists etc.

bures:
Yes. This feature would be nice indeed. The same for a template.

shwaydogg:
It's nice that duplicate template is now a feature in #344 but duplicate a campaign would also be nice... 

Or am I missing a UX step that would allow this?

ansgarsmith:
Same here: copy campaign would be really effective to have :-)

ateuber:
We could also need the possibility to duplicate a campaign.

forgeOneHQ:
It is already implemented in v2. Please install v2.

PR #344, opened 2017-10-16
Adds duplicate Template

brentdur:
This adds a duplicate button the edit template page. Pressing this button gets the template by id and creates a new template with retrieved data, appending "Copy" to the template name and redirecting the user to the edit page for the new template.

For issue #247

### DIFF
--- a/lib/models/templates.js
+++ b/lib/models/templates.js
@@ -61,15 +61,17 @@ module.exports.create = (template, callback) => {
     let values = [name];
 
     Object.keys(template).forEach(key => {
-        let value = template[key].trim();
+        let value = template[key];
         key = tools.toDbKey(key);
+        if (!allowedKeys.includes(key)) {
+            return;
+        }
+        value = value.trim();
         if (key === 'description') {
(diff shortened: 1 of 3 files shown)


witzig:
Hi @brentdur, thanks for this PR.

It seems to me the changes in `module.exports.create` are unnecessary and your `return` statement in the 'forEach' loop also exits the function. Should we revert this?

brentdur:
@witzig Hi.

The changes in `module.exports.create` were done due to an error if a template with a `id` field was passed into create. `id` is not an allowedKey and thus shouldn't be processed, yet `.trim()` was still being called which was causing an error as id isn't a string. (Believe it was `id`, this PR was some time ago so I can't remember exactly which field)

The `return` is inside the `forEach` which is a closed function, so I believe return will exit the `forEach` function (continuing to the next value) but will remain inside the `module.exports.create` function. As it isn't an actual loop statement I think `continue` would actually fail.

witzig:
Thanks @brentdur!

PR #348, opened 2017-11-08
Add API support for GET /api/subscriptions/:listId

gkrol-zermelo:
I needed a way to retrieve all subscriptions to a list, in order to synchronize mailtrain with our own administration. I'm not that happy with the fact custom fields get weird generated names in the JSON, but in practice it's not a problem as those are easily copy-pasted into any code that uses them.

### DIFF
--- a/routes/api.js
+++ b/routes/api.js
@@ -315,6 +315,39 @@ router.post('/delete/:listId', (req, res) => {
     });
 });
 
+router.get('/subscriptions/:listId', (req, res) => {
+    let start = parseInt(req.query.start || 0, 10);
+    let limit = parseInt(req.query.limit || 10000, 10);
+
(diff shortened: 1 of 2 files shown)


afilp:
Hello, can this be merged please? It seems to be OK, thanks!

witzig:
Thanks @gkrol-zermelo !

afilp:
@witzig I feel a little strange too that there is a random naming in the names of custom fields.

gkrol-zermelo:
I found this branch, which does seem to have nicer names for the custom fields: https://github.com/afanjul/mailtrain
We might be able to borrow some code from there.

afanjul:
Didn't I do a pull request of that feature? I thought I did, but anyway if you need that I do a pull request for some commit, just paste here a link to the commit... (sorry im so busy these days)

PR #349, opened 2017-11-08
Add support for passport-ldapauth

gkrol-zermelo:
passport-ldapjs wasn't working for me, as the DN it generates for the initial bind somehow doesn't work at all with our SAMBA AD. To fix this I added support for the (much more popular) passport-ldapauth library. That one works like a charm.

I'm not sure Mailtrain should support multiple ldap libraries, so I would propose removing passport-ldapjs support. Then we probably also should modify the configuration variables to be as close as possible to the passport-ldapauth ones.

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -125,6 +125,22 @@ passwordresetlink=""
 bindUser=""
 bindPassword=""
 
+[ldapauth]
+# Alternative LDAP implementation using the more popular passport-ldapauth library.
+enabled=false
+host="localhost"
+port=389
+# Subtree in which the searchrequest for the user is done
+baseDN="ou=users,dc=company"
+# What whe are searching for. This should return a single user.
+filter="(|(sAMAccountName={{username}})(mail={{username}}))"
+# Username field in LDAP, used to identify the user in Mailtrain (uid/cn/username/sAMAccountName)
+uidTag="sAMAccountName"
+passwordresetlink=""
+# Credentials for the initial search operation (final bind DN will be exactly as specified)
+bindUser="name@company.net"
+bindPassword="mySecretPassword"
+
 [postfixbounce]
 # Enable to allow writing Postfix bounce log to Mailtrain listener
 # If enabled, tail mail.log to Mailtrain with the following command:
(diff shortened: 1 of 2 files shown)


witzig:
> [...] I added support for the (much more popular) passport-ldapauth library. [...] I'm not sure Mailtrain should support multiple ldap libraries, so I would propose removing passport-ldapjs support.

Seems reasonable to me. @larrabee, @nebulade, @MelwinKfr, as far as I'm aware you've all contributed to the current ldap implementation. What's your take on this?

MelwinKfr:
It seems to be a good idea, my contribution is useless using passport-ldapauth since it already implement a way to modify the bind request.

nebulade:
This is probably a good idea to replace the passport-ldapjs support with this if only because passport-ldapauth seems to be a lot more actively developed than the other alternatives. From what I can tell it is mostly feature equal for the mailtrain use-case.

alen12345:
@gkrol-zermelo 'passport-ldapauth' is not installed by default; we should add it to the Dockerfile as for 'passport-ldapjs'

In addition:
` (config.ldapauth.enabled && LdapAuthStrategy) {

TypeError: Cannot read property 'enabled' of undefined
`

Can you replicate this?

gkrol-zermelo:
@alen12345 This pull request has already been merged, so please create new issues for this.

I'm not using docker, so feel free to add passport-ldapauth to the dockerfile and send a pull request.

About the missing config.ldapauth: I assume that's because the entire section is missing. It should probably be tested for presence like (config.ldapauth && config.ldapauth.enabled && ...). I'm actually not a JS programmer so that may need to be slightly different.

alen12345:
@gkrol-zermelo unfortunately your pull request, which to me is very useful, totally breaks the docker support for mailtrain. There must be something going on when copying the files as you say.

I will add the package to the Dockerfile, check the problem with the Javascript, open a new issue and reference a pull request as suggested.

Thank you for this new feature by the way!

alen12345:
Your pull request does not break docker support, I am sorry to have said that; the problem is that the new default.toml is not copied over to the container upon rebuild...

I have added the new dependencies to the Dockerfile and now issuing a new pull request.

PR #377, opened 2018-02-06
Allow CORS headers to set in `api/*`

ashleytqy:
Currently, CORS headers are only set in `routes/subscription.js`, so when a call is made to `/api/subscribe/:list_id`, there will be a "No 'Access-Control-Allow-Origin' header" error.

This PR moves the logic of the CORS Options into `lib/tools.js` so that it can be used in both `routes/api.js` and `routes/subscriptions.js`.

Do let me know if `lib/tools.js` is a suitable place for this function, and thanks @brentdur for the help!

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -14,7 +14,6 @@ let _ = require('./translate')._;
 let util = require('util');
 let createDOMPurify = require('dompurify');
 let htmlToText = require('html-to-text');
-let config = require('config');
 
 let blockedUsers = ['abuse', 'admin', 'billing', 'compliance', 'devnull', 'dns', 'ftp', 'hostmaster', 'inoc', 'ispfeedback', 'ispsupport', 'listrequest', 'list', 'maildaemon', 'noc', 'noreply', 'noreply', 'null', 'phish', 'phishing', 'postmaster', 'privacy', 'registrar', 'root', 'security', 'spam', 'support', 'sysadmin', 'tech', 'undisclosedrecipients', 'unsubscribe', 'usenet', 'uucp', 'webmaster', 'www'];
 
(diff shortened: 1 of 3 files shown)


bures:
@ashleytqy, the code looks OK. However, I was wondering why is this needed. If I understand right, it allows the use of the API from web-browsers. Right? If so, I'm not completely sure this should be allowed by default. Could you please describe your use-case?

bures:
Doesn't the use of API calls require you to send the API key to the web-browser? If so, are you aware that this is a serious security flaw?

If you care only for subscribing users, you can use the same method that the subscription widget is using. You don't have to use the widget directly, you just insert in the page your piece of code that makes a similar request on the server.

ashleytqy:
Yup totally understand the security issues. However, if I wanted to make an API call from the server to our Mailtrain server, I'd still need the CORS headers to be set on the Mailtrain server for the API call to go through.

bures:
Nope. If you do the API call from a server, the CORS headers don't matter. They are used only when calling from a web-browser.

ashleytqy:
Ohh gotcha! Re: the API key being a security issue, what if it was set in an `ENV` variable?

brentdur:
ENV variables don't apply to frontend web applications. Would make sense for server -> server calling case though (and would be the best way to mitigate the security concern)

If I understand correctly, I think using the widget endpoint makes the most sense for frontend registration outside of the [widget](https://github.com/Mailtrain-org/mailtrain/blob/master/views/subscription/widget-subscribe.hbs). It could be worth while to document this on the `users/api` page to make it clearer in the future as well

bures:
There seems to be no further discussion to this pull request. The conclusion seems to be that it would introduce security issues. As such, I'm closing this request without merging.

PR #383, opened 2018-02-13
Added ldapauth support to docker

alen12345:
Added support in Docker for pull request #349

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM node:8.6
 COPY ./package.json ./app/
 WORKDIR /app/
 ENV NODE_ENV production
-RUN npm install --no-progress --production && npm install --no-progress passport-ldapjs
+RUN npm install --no-progress --production && npm install --no-progress passport-ldapjs passport-ldapauth
 # Later, copy the app files. That improves development speed as buiding the Docker image will not have 
 # to download and install all the NPM dependencies every time there's a change in the source code
 COPY . /app


saschafoerster:
I tried to install via Docker, but I get the same error messages as you: 
> TypeError: Cannot read property 'enabled' of undefined

When I replace the passport.js with an older version without the ldap-changes, everything works. 
https://github.com/Mailtrain-org/mailtrain/commit/65923c0c050f1306e15791af8a769d5affe0d697#diff-567284bb08dc023ac92bad5d592343fe

alen12345:
@saschafoerster the problem is that, if you pull the new commits, you don't copy over to the container the new default.toml file. Can you please check this fact?

saschafoerster:
Okay, this was the solution to the problem:
I had to add another line to the docker-composer.override.yml
Before:
```
    volumes:
      - "/home/mailtrain/config/production.toml:/app/config/production.toml"
```
after
```
    volumes:
#      - "/var/run/mysqld/mysqld.sock:/var/run/mysqld/mysqld.sock"
#      - "/var/run/redis/redis.sock:/var/run/redis/redis.sock
      - "/home/mailtrain/config/default.toml:/app/config/default.toml"
      - "/home/mailtrain/config/production.toml:/app/config/production.toml"
```
So no more of this error:
```
TypeError: Cannot read property 'enabled' of undefined
    at Object.<anonymous> (/app/lib/passport.js:129:27)
    at Module._compile (module.js:624:30)
    at Object.Module._extensions..js (module.js:635:10)
    at Module.load (module.js:545:32)
    at tryModuleLoad (module.js:508:12)
    at Function.Module._load (module.js:500:3)
    at Module.require (module.js:568:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/app/app.js:20:18)
    at Module._compile (module.js:624:30)
info LANG Loaded language file for de_DE_bonndigital
/app/lib/passport.js:129
} else if (config.ldapauth.enabled && LdapAuthStrategy) {
```
Thanks @alen12345!

PR #385, opened 2018-02-23
Fixed a bug in percentage calculation of mail transmission result

cold00water:
Campaign result percentage is strange.
ex)
total delivered: 3
bounced: 1 (50%)

This pattern should be 33%, but this result was due to an internal calculation error.

### DIFF
--- a/routes/campaigns.js
+++ b/routes/campaigns.js
@@ -328,14 +328,17 @@ router.get('/view/:id', passport.csrfProtection, (req, res) => {
 
             campaign.isScheduled = campaign.scheduled && campaign.scheduled > new Date();
 
+            // save total delivered count
+            campaing.orgdelivered = campaign.delivered
+
             // show only messages that weren't bounced as delivered
             campaign.delivered = campaign.delivered - campaign.bounced;
 
-            campaign.openRate = campaign.delivered ? Math.round((campaign.opened / campaign.delivered) * 10000) / 100 : 0;
-            campaign.clicksRate = campaign.delivered ? Math.round((campaign.clicks / campaign.delivered) * 10000) / 100 : 0;
-            campaign.bounceRate = campaign.delivered ? Math.round((campaign.bounced / campaign.delivered) * 10000) / 100 : 0;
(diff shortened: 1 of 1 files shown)


bures:
Where is the "Total delivered" field? I see only "Delivered", which is then consistent with the other fields (as it shows already the total without the bounces). Could you please send a screenshot?
Thanks.

cold00water:
Sorry I can not convey it well.

This is screen shot.
The bounce value is 2 (200%).
Is not this strange?

screenshot-2018-2-27 mailtrain

bures:
Oh yes. Now I understand. The bounces are indeed a problem. 

I'm thinking what would be the right interpretation of "Complaints", "Unsubscribed", "Opened" and "Clicked". The "Delivered" field shows that #subscribers - #bounces. "Complaints", "Unsubscribed", "Opened" and "Clicked" are relevant only when the email does not bounce. As such, I could imagine that these percentages would be from the number of really delivered emails (i.e. without the bounces) as opposed calculating it from the number of all sent emails.

When thinking about it further, it even does not make sense to show bounces as a percentage. It would be much better to show them as a number - similar to Delivered and Blacklisted. Then if the "Complaints", ... stay as they are, they make sense, because those are computed against the "Delivered".

What do you think? Does it make sense?
Thanks.

bures:
@cold00water, please any reaction to my previous comment? I would be happy to have this improvement. Just when doing it it, I thought it would be good to do it overall right. Would you like to extend your pull request to contain those things suggested?

routes/campaigns.js
```diff
@@ -328,14 +328,17 @@ router.get('/view/:id', passport.csrfProtection, (req, res) => {
 
             campaign.isScheduled = campaign.scheduled && campaign.scheduled > new Date();
 
+            // save total delivered count
+            campaing.orgdelivered = campaign.delivered
```
jc1arke: @cold00water hey dude, think you missed the spelling mistake there :) `campaing` should be `campaign`

PR #402, opened 2018-03-24
Fix translations typo and double message

rrooij:
The Spanish error about the missing message ID appeared twice and
generated an error when converting the translation file.

Another fix is the typos that were still present in the msgid's. The mo files have been recompiled.


bures:
Excellent. Thanks.

PR #408, opened 2018-04-10
[LANG] 94% Italian language translation - left some techy strings

giorgioma:
Title is pretty self-explanatory  I guess


bures:
Excellent. Thanks.

Issue #403, opened 2018-03-28
Question: Triggered Campaigns and Segments


Roost:
After many hours setting up and playing with segments and automation, I realised that segments cannot be used to send triggered campaigns. 

Firstly, is this in fact true or is there something I am missing or not understanding?

Secondly, if this is indeed correct, can anyone suggest an approach that will allow automated campaigns to send emails only to qualifying addresses. Since we update users via the API using the merge tags, we have granular segment management in place. 

From the looks of things, currently the only approach would be moving users between lists based on their segmentation, but I worry that this may mean a user could receive duplicate emails from moving into and out of a list multiple times. Also, this would mean the server handling the API updates would need to ping the Mailtrain API repeatedly on a cron like basis to ensure the users are correctly moved between lists- this seems very inefficient.

I love Mailtrain, but really need a mechanism to handle automation based on small sub segments of the larger list. Suggestions welcome.

vilius-rugaitis:
I am also experiencing this behaviour. Are there any chances, that this feature will be available soon?

bures:
I'm afraid it will take at least a few months. I'm currently working on complete rewrite of Mailtrain. As such, I don't want to invest time in extending the existing version. The support for automation is one of the lesser priority points on my TODO list for the new Mailtrain, but it should be appear there too. Once the automation is implemented for the new Mailtrain, it should contain also the possibility to work with segments.

jonasduever:
@Roost @bures #412 implements this, but probably needs some more testing (it works in my environment, though). Feel free to leave any remarks or comments, especially about bugs you may experience, thanks!

notz:
Seems that this feature is now v1 but not in v2 beta. Any chance to get in v2?

dlecan:
My 2 cents about this feature (Mailtrain v2 beta):
- we set up several triggered campaigns based on segments a few months ago
- then we realized they didn't work (the current issue :grin: ), emails weren't triggered with segments
- BUT we didn't do anything else
- Today, our Mailtrain server was restarted ...

👉 Hundreds of emails were sent 😱 all these which should have been sent by triggered campaigns with segments we've set up

So, it is untrue to write that triggered campaigns don't work with segments: they work only when the Mailtrain server is started :wink:!

So most of the things seem to be already coded for this to work.

@Mailtrain developers: where should I look at to enable triggered campaign with segments "on the fly"?

Thank you

forgeOneHQ:
[...]

PR #412, opened 2018-04-25
Support for Triggers on Segments

jonasduever:
Since I needed the functionality mentioned in #403 and there is no plan to implement it in v1.x of Mailtrain, I implemented it.

The implementation adds segments to the triggers frontend and includes the necessary modifications of the trigger SQL queries. It needs a slightly modified database structure (added 'segment' columns), so an SQL upgrade script has been added to the setup/sql folder.

### DIFF
--- a/lib/models/triggers.js
+++ b/lib/models/triggers.js
@@ -1,8 +1,10 @@
 'use strict';
 
+let log = require('npmlog');
 let tools = require('../tools');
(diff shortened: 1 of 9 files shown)


bures:
Looks good. I just would like to test it a bit before merging it.

bures:
Thanks. I bumped the version of sql update and merged it.

PR #413, opened 2018-04-26
Fix spelling in email field help block.

dougal:
Changed

> This address is used for account recovery in case you loose your password

to

> This address is used for account recovery in case you lose your password


bures:
Thanks.

PR #426, opened 2018-05-23
Fix newsletter signup link in README

wesgarrison:
I replaced the link with the one from the mailtrain.org homepage:

image


bures:
Thanks.

Issue #422, opened 2018-05-16
Allow to disable "List-Unsubscribe" header


njam:
I'm using Mailtrain for sending promotional campaigns to my website's users. Now my website has its own *unsubscribe* feature, and I would like to manage the list there primarily. I'm importing a field `unsubscribe_link` with the list, which I include in the email. But Mailtrain still includes its own unsubscribe link the the `List-Unsubscribe` header.

Basically I would like my users to never interact with the Mailtrain website directly.
What do you think about allowing to disable the `List-Unsubscribe` header?

The header is added in `services/sender.js :: formatMessage()`.

(another solution could be to make the header configurable through a merge tag from the list, but I guess that would be quite complicated.)

njam:
Would this get accepted if I would introduce a global configuration option to disable `List-Unsubscribe` headers?

bures:
I would be fine with making this configurable on the list level (i.e. lists table in DB) or via the global conf. option.

Just note that ain't won't likely keep the global conf option in the next major version of Mailtrain (which won't be around for at least several months). I'll likely reflect it on the list level.

njam:
@bures if the right place is on the *list level*, then I will add it there. So you prefer if I add a column `disable_listunsubscribe` to the `lists` table, right?

bures:
Yes. That would be great. Thanks.

njam:
This is implemented and merged.

@bures What would be the procedure to get this change added to the next release? Is there any new release planned?

PR #434, opened 2018-05-29
Add option on list to disable "List-Unsubscribe" header

njam:
Addressing #422

I've added a mysql column `lists.listunsubscribe_disabled`, defaulting to "0" and a corresponding field in the model. I've done some acceptance testing and have confirmed that it behaves as expected.

I'm not sure about the description and placement of the option. For now it's only visible on the *edit* page but not on the *create* page of a list (with the reasoning that this is a rather edge case configuration). Please let me know if there's a better way to do it.

@bures wdyt?

### DIFF
--- a/lib/models/lists.js
+++ b/lib/models/lists.js
@@ -19,7 +19,7 @@ const UnsubscriptionMode = {
 
 module.exports.UnsubscriptionMode = UnsubscriptionMode;
 
-let allowedKeys = ['description', 'default_form', 'public_subscribe', 'unsubscription_mode'];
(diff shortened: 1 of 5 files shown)


njam:
@bures is this what you had in mind? Please let me know what should be changed!

bures:
Looks good. Do you have any idea why the tests are failing? Did you try to run them locally? Thanks.

njam:
@bures after #436 and adding a missing semicolon it's green now!
I ran tests locally and also did some acceptance testing locally.

bures:
Looks good. Thanks.

PR #444, opened 2018-06-21
LDAP SSL Support

Addy90:
Please find submitted a suggestion to enable LDAPS without destroying LDAP compatibility.
For enabling this, the configuration file was changed from host and port to url so that the previously hardcoded 'ldap://' is now in the scope of the user.
One difficulty of ldaps:// connectivity is that nodejs does not use the system-wide installed CAs but (like Firefox does) only supports Mozilla provided certificates (see [tls_tls_createsecurecontext_options](https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options)). This is why an optional ca config item was added for providing a self-signed certificate or ca (not needed with ldap://).
What is currently still unsupported is StartTLS as this requires a different stream processing of the connection. [ldap-js supports this](http://ldapjs.org/client.html#starttls), but this is more complicated and LDAP servers that support (or require) StartTLS are easily configured to support standalone SSL, as well, so not supporting StartTLS is no real blocker.

I hope you acknowledge these changes! I tested for myself and was able to successfully login via SSL.

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -114,8 +114,7 @@ host="0.0.0.0"
 [ldap]
 # enable to use ldap user backend
 enabled=false
-host="localhost"
-port=3002
+url="ldap://localhost:3002"
 baseDN="ou=users,dc=company"
 filter="(|(username={{username}})(mail={{username}}))"
 #Username field in LDAP (uid/cn/username)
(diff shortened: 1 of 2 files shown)


bures:
Thanks.

Issue #262, opened 2017-06-22
Enabling https


tuck1s:
Simple mods to enable https access, taking standard `install.sh` setup as a starting point, to permit secure running on public servers. Ideally for both UI and webhooks receptors.

tuck1s:
There's a simple strawman version in here https://github.com/tuck1s/mailtrain

witzig:
@tuck1s, nice work regarding [your mods for enabling https access](https://github.com/tuck1s/mailtrain/commit/69158f433955a714eabd2dd9e0eb338d6f458bbd).

Generally, I think the SSL stuff is better handled with Nginx or Apache. In case someone's looking for a tutorial, here's a great [guide for Nginx](https://www.digitalocean.com/community/tutorials/how-to-secure-nginx-with-let-s-encrypt-on-ubuntu-16-04) using Let's Encrypt.

christiangenco:
Is there any progress on enabling SSL for hosted mailtrain instances? I looked through [the guide for adding SSL to nginx](https://www.digitalocean.com/community/tutorials/how-to-secure-nginx-with-let-s-encrypt-on-ubuntu-16-04) @witzig posted, but I don't think mailtrain is using nginx on my instance (I don't have a `/etc/nginx/` directory)

Wooza42:
It would be great if mailtrain supports https. I did not manage to get it work with apache2 and mod_proxy, it always says: The proxy server received an invalid response from an upstream server.
The proxy server could not handle the request GET /.

MiBaDK:
And if you are too lazy, like me, to mess around with SSL config then just use CloudFlare and force SSL under crypto.

![image](https://user-images.githubusercontent.com/27769361/41212292-64f74a74-6d6f-11e8-8539-d7c2c75bb3b4.png)

projectsfor:
Installed mailtrain on subdomain, it's working perfectly until https=true.
SSL generated with certbot and after https=true I have not working **http**://subdomain.domain.com and **https**://subdomain.domain.com loading content from https://domain.com
also by **https**://ip:port I'm getting not protected mailtrain content
by **http**://ip:port - site unavailable
conf file: 

> https=true
> # HTTPS certificate file name
> cert="/etc/letsencrypt/live/mailtrain.domain.com/fullchain.pem" (or cert.pem)
> #cert="cert.pem"
> # HTTPS certificate private key file name
[...]

PR #445, opened 2018-06-21
Enabling HTTPS server side support

Addy90:
this permits [https](https://nodejs.org/api/https.html#https_https_createserver_options_requestlistener) as asked in #262 
it can be enabled or disabled by the new config options.
it is possible to serve certificate, keyfile, ca-certificate if needed and dhparams if needed.
I tested with and without dhparams and with and without https enabled, it worked as assumed.


bures:
Thanks.

Issue #399, opened 2018-03-20
Question: NOT operators in segment rules


Ridingst:
Hi,

Great work all involved, I have been loving this project!

When building segment rules is it possible to specify a "NOT" operator or "must be null"?

I've inserted (via API) lots new users with a null value into a custom field that was previously yes/no. As a result I can no longer segment these users.

I could write some custom scripts to pull from the API and correct the problem but a NULL or NOT selector feels like something that could solve the problem much quicker.

PR #449, opened 2018-07-02
Add the possibility to create a negated rule on segments

humancopy:
Hey,

This adds the option to create negated rules for segments.

So for example you can create a segment on all signup countries which are not France.

My first idea was to actually create a more user friendly rule setup, so it's possible to choose from options like:

Equals
Not equals
Starts with
Ends with
Contains

Let me know if this is something you interested in and I'll make a pull request for it... Thanks 😄


humancopy:
#399

bures:
Thanks.

PR #450, opened 2018-07-11
GrapeJS Update to latest version

Addy90:
Because of my need of the new "clearProperties" feature of GrapeJS, I made an upgrade on my own and enabled that feature. I hope that it is fitting and has no side effects. Please review before merging! I hope that everything works...


Addy90:
Looks like the "Test Newsletter" Button is not working - as this one does not send correct E-Mails without the Markup Tags evaluated, I don't think this is a serious problem, but if, please help with this one!

Addy90:
The "Test Newsletter" Button is working again.
Also, images were defective, but they work now, too.
For me, everything looks good with the update.

bures:
Thanks.

Addy90:
Thank you!
Concerning the "hack" I needed for the Test button to work: Somehow, GrapeJS has a bug in it's code. When I call "setContent" with the original HTML content for the send-test form, you get an error about length is undefined from GrapeJS. The problem is that the elements to be added are referenced by array indices, for normal HTML elements, this works for the children, but form elements only show the form elements, not the other elements. The underlying code awaiting other inputs stops working then.
I could fix this behavior by not adding a form html element but a string of the content. I had to change other references for the code but after that, it worked.
Then I had a last problem, somehow the content is not correctly removed from the modal dialog when closing, so additional calls of setContent do not work then, except after several tries (like every 3rd). Bad fix was to just try-catch in a loop. I don't like that behavior but GrapeJS modal dialog seems to have problems with complex content. After all, it workes so far for this version now and one can use the new "clearProperties" feature now.

PR #453, opened 2018-08-01
Add custom unsubscribe url option in a campaign

raphaelhittich:
https://github.com/Mailtrain-org/mailtrain/pull/451 should be pulled first, otherwise there could be conflicts.

### DIFF
--- a/services/sender.js
+++ b/services/sender.js
@@ -381,7 +381,7 @@ function formatMessage(message, callback) {
 
                                 let listUnsubscribe = null;
                                 if (!list.listunsubscribeDisabled) {
-                                    listUnsubscribe = url.resolve(configItems.serviceUrl, '/subscription/' + list.cid + '/unsubscribe/' + message.subscription.cid);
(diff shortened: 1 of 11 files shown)


bures:
Thanks.

services/sender.js
```diff
@@ -381,7 +381,7 @@ function formatMessage(message, callback) {
 
                                 let listUnsubscribe = null;
                                 if (!list.listunsubscribeDisabled) {
-                                    listUnsubscribe = url.resolve(configItems.serviceUrl, '/subscription/' + list.cid + '/unsubscribe/' + message.subscription.cid);
+                                    listUnsubscribe = campaign.unsubscribe ? tools.formatMessage(configItems.serviceUrl, campaign, list, message.subscription, campaign.unsubscribe) : url.resolve(configItems.serviceUrl, '/subscription/' + list.cid + '/unsubscribe/' + message.subscription.cid);
```
njam: @raphaelhittich does this mean the "unsubscribe" setting can contain merge tags?

raphaelhittich: Exactly, for example:

https://example.com/unsubscribe/[MERGE_ID]

njam: Nice!

PR #457, opened 2018-08-07
add passport-ldapauth to dependencies

Addy90:
when "npm install --production" is called, every time, passport-ldapauth is removed from local repository, with adding the dependency to package.json, the behavior is as expected (see production.yml config for ldap config)


bures:
It seems I can't merge it my phone. I'll merge it once I get to my laptop.

Addy90:
No problem, thank you for your fast response!

Issue #103, opened 2016-10-24
Multiple Copies/Domains?


stevenmcastano:
I really want to be able to run this for multiple different domains and I installed it on a central email server I used, but I'm noticing now that basically everything is one large running copy.

So for instance, my default email is set to "do-not-reply@myemailserviceibuilt.com", so all of the confirmation emails and such come out from that address and even after creating new users in the database, they don't have their own email server settings, so I can really use a single running copy of this to cover multiple uses.

Also, things like the "view this email in your browser" link, can't really be set by default to another domain.... etc...

Would it be possible to just create multiple .toml files like domain1.toml, domain2.toml, etc... and just have them all point to different databases so I could have different settings, but still run from the same working directory? Or are there files on disk that will contain urls and such of where it was built?

Then if I do have a domain1.toml file, could I just create multiple upstart/systemd scripts to run the multiple copies on multiple ports at the same time?

andris9:
You can define config file with NODE_APP_INSTANCE environment variable, so for example if you have set NODE_APP_INSTANCE=1 and NODE_ENV=production then the config file to load (in addition to default.toml) would be production-1.toml. I haven't actually tried it but this is what the config package [wiki says](https://github.com/lorenwest/node-config/wiki/Multiple-Node-Instances) that Mailtrain is using.

balping:
I confirm, the above method works.

PR #466, opened 2018-09-08
Ignore more config files

balping:
To run multiple instances, you need to create files like `production-1.toml`, `production-1.toml`, etc. as described in #103 

These config files should be ignored too.


bures:
Thanks

PR #471, opened 2018-09-30
API: Add the possibility to fetch the details for a specific subscription

humancopy:
This commit adds a /subscription/:listId/:email route that can be used to fetch the details for a specific subscription ...


humancopy:
Hey guys any reason why this is not being merged? It's a very useful feature to be able to integrate with external systems, we use it all the time 🤓

PR #480, opened 2018-10-23
API: Add the possibility to change the email address of an existing list subscriber

ateuber:
This commit adds a `/changeemail/:listId` route that can be used to change the email address of an existing list subscriber. It is checked that the new email address is valid and not on the blacklist.


bures:
Thanks

PR #482, opened 2018-10-31
Custom field description

ateuber:
Adds a description textfield to custom fields of lists and makes it visible on the sites to subscribe or manage own subscriptions.


bures:
Thanks.

Issue #532, opened 2019-02-16
DB migration to v2 fails


priethor:
After trying my v2 instance and making sure it's properly setup (the scripts provided don't work in Ubuntu 16.04, should I send a PR with 16.04 support?), I tried to run it against my v1 database and the migration failed, leaving this traces in the system log:

> Feb 17 00:59:21 xxx.xxx.xxx systemd[1]: Started Mailtrain server.
> Feb 17 00:59:24 xxx.xxx.xxx node[31933]: info Using local auth
> Feb 17 00:59:25 xxx.xxx.xxx node[31933]: info sql Update 31 applied
> Feb 17 00:59:25 xxx.xxx.xxx node[31933]: info sql Update 32 applied
> Feb 17 00:59:25 xxx.xxx.xxx node[31933]: info sql Update 33 applied
> Feb 17 00:59:25 xxx.xxx.xxx node[31933]: info sql Database check completed
> Feb 17 01:05:48 xxx.xxx.xxx node[31933]: migration file "20170506102634_v1_to_v2.js" failed
> Feb 17 01:05:48 xxx.xxx.xxx node[31933]: migration failed with error: Unknown rule for column email with field type string
> Feb 17 01:05:48 xxx.xxx.xxx node[31933]: Unhandled rejection Error: Unknown rule for column email with field type string
> Feb 17 01:05:48 xxx.xxx.xxx node[31933]:     at migrateSegments (/opt/mailtrain/server/setup/knex/migrations/20170506102634_v1_to_v2.js:635:27)

priethor:
A bit more info on how I came to this error:
- Populate the brand new mailtrain DB from a v1 mysql dump with `mysql mailtrain < mailtrain.sql`
- Started mailtrain service. It applies pending SQL updates from `mailtrain/server/setup/sql`:

> Feb 17 00:59:21 xxx.xxx.xxx systemd[1]: Started Mailtrain server.
> Feb 17 00:59:24 xxx.xxx.xxx node[31933]: info Using local auth
> Feb 17 00:59:25 xxx.xxx.xxx node[31933]: info sql Update 31 applied
> Feb 17 00:59:25 xxx.xxx.xxx node[31933]: info sql Update 32 applied
> Feb 17 00:59:25 xxx.xxx.xxx node[31933]: info sql Update 33 applied
> Feb 17 00:59:25 28810.s.t4vps.eu node[31933]: info sql Database check completed
- The service stops there. I 
-

[...]

PR #534, opened 2019-02-16
Support for string based segment rules....

priethor:
...such as email fields.


priethor:
Closes #532

bures:
Perfect! Thanks.

Issue #535, opened 2019-02-17
Migrated text based segment rules don't contain the "type" setting


priethor:
Following #532, where text-based rules crashed the migration script, the migration now finishes without errors but text-based rules are not migrated properly.

In v1, the rules where simpler to the extend they had only a rule (field) and a value, but now we can also define directly a type such as "equals to", "match", etc. The migrated text rules don't contain any type, only the column and the value.

priethor:
As in v1 the value could contain wildcards such as %, I guess the default type added in migrations should be `Match (with SQL LIKE)`, right?

bures:
Yes. I guess so. Thanks.

PR #537, opened 2019-02-17
Adds 'type' property to migration of text-based segment rules

priethor:
In v2, text-based segment rules need a `type` property. As in v1 the value could contain `%` wildcards, the default type for migrated rules should be `like` to support them.


priethor:
Closes #535

bures:
Excellent. Thanks.

Issue #563, opened 2019-03-20
v2 - Current namespace filtering in campaigns


priethor:
As mentioned in #562, in order to avoid permission issues, it would be a huge UX improvement to not be able to assign to a campaign (or template) elements not belonging to the campaign nampesaces.

We have a few suggestions:

- Directly filtering the lists, templates and send configurations available to chose in the campaign by the selected nampesace.
- When selecting the campaign namespace, adding a checkbox to enable/disable the automatic filtering (unchecked would work as it is right now, checking it will work like in the previous option)
- Adding a checkbox to all chosable elements (lists, send configuration, template) to enable the filtering by the current campaign namespace.

Which one do you prefer, @bures ?

priethor:
The 3rd option is implemented at the moment in PR #564

Issue #543, opened 2019-02-19
Feature request - Namespace filtering


priethor:
Now that v2 supports namespaces and the application can be "partitioned" allowing multitenancy, it would be really cool to be able to filter by an "active namespace" provided the user has access to more than one.

For example, in our Mailtrain instance we have campaign managers having access to the root namespace or a set of namespaces, but our typical use case involves "concentrating" on a namespace and jumping between the content available only in that namespace, as in:
1. Manage list(s) assigned to namespace A
2. Manage templates and/or campaigns assigned to namespace A
3. Go back to 1 if required
4. Manage list(s) assigned to namespace B
5. Manage templates and/or campaigns assigned to namespace A
6. Go back to 4 if required
7. Etc.

It would be very helpful to have a dropdown in the top bar to select the "active namespace", so that all content is filtered by that namespace as if the user didn't have access to other namespaces.

bures:
I agree. Good point. I had something similar in mind too. I guess, we can agree on having some dropdown with namespaces in the top bar. 

Would it be useful to have also a checkbox "Include sub-namespaces"? Or is it something you don't need in your use-case? I'm asking because implementing the children will be some complication, which is however worth it, if this is a needed use-case.

Thanks.

priethor:
Honestly I hadn't even considered it, as I was assuming it would work in the same way as the user roles do, which they do include the children. For example, if a user has Campaign Admin role in the root namespace, he sees everything belonging to its subtree, so I guess this dropdown should be consistent with that behaviour.

In my use case, we are just starting with v2 and still arranging our content hierarchically according to the namespaces, so I really can't tell whether single node or subtree is better, but with just a few days in v2 after a year with v1, I'm only certain that it needs to be consistent with the roles or it can become messy and difficult to understand.

bures:
I see. A bit of the issue is that roles can be configured whether to include the sub-namespace content or not. It's more like a coincidence that Campaign admin is configured to include sub-namespaces. You can see it in https://github.com/Mailtrain-org/mailtrain/blob/0c3510d62695e05b58f83da447d6b9564ab12300/server/config/default.yaml#L259

The `children` section determines what permissions there to entities in children namespaces. You can thus theoretically have larger permissions to the namespace than you have to the children. Though I had no use case for this so far.

[...]

PR #564, opened 2019-03-20
Bugs and added features

Charlo270398:
Hello, I have fixed some bugs and added features. You can depreciate the initial commit "works" (I needed it to run Mailtrain on my own computer).


bures:
Excellent. I'll take a look at it.

bures:
Thanks.

bures:
I'm merging this PR and polishing the code. However, I don't fully understand the idea behind the "Work with campaign's namespace". In my view, the namespace selection should be somewhere in the top-right corner and it should apply to everything (not just campaigns). 

Thus, I took over the features that pertain to visibility of user settings, the confirmation dialog for send, the adjustment of position of the "Overrride" boxes, etc. However, I left out the namespace filtering. 

Please check - https://github.com/Mailtrain-org/mailtrain/tree/development-pr-564

If that work, I'll merge it to deployment. Also, if you like to implement the universal namespace filter (i.e. the one in the top-right corner or so), I would appreciate it. In that case, please base it on the development-pr-564 branch.

Thanks.

priethor:
> However, I don't fully understand the idea behind the "Work with campaign's namespace". In my view, the namespace selection should be somewhere in the top-right corner and it should apply to everything (not just campaigns).

The idea is to be able to filter, appart from with the top right global filtering as talked in #543, send configurations and/or lists to show only those the campaign has access to, always according to its namespace (see #563) in order to avoid misconfiguration. In this PR, the most granular option is implemented, which is probably too atomic, but always as a starting point for #563 and awaiting your input.

bures:
I see. A bit of the problem is that you cannot say whether the campaign has access to the list, send-configuration, etc. unless you have a particular user. The reason for that is that you can individually share an entity (i.e. list, send-configuration, etc.) with someone. They don't have to be in the same namespace or whatever.

Also, I would rather keep the UX rule that whatever fields are shown in the entity settings, they somehow reflect what is persisted in the DB. The checkboxes for freezing the namespace are just view filter. As such, they should be kept somehow separate from the settings panel. I could even imagine that this could be another option in the right-hand corner. 

We could have options like:
- show only entities (including namespaces) only from a particular namespace and its children
- show only entities (including namespaces) only from user's namespace and its children
  - in this respect we could also introduce to a user a checkbox to automatically pre-select this option

Plus we could have a checkbox somewhere in the dropdown from this top-right namespace selector where we offer a checkbox "In entity settings, filter dependent entities by the main entity". This would have the same semantics as what you introduced to campaign settings, only it would be placed as part of the entity filtering control (where I believe it belongs).

What do you think please?

And please, let me tell you I really appreciate your readiness to contribute.

bures:
BTW, I merged the cleaned up PR in the development branch. I have also fixed line endings of some older files (so that we have on LF in git). Please, if you continue on this, checkout the development branch and build it on top of it. I'll delete the development-pr-564 branch now.

Issue #352, opened 2017-11-20
How to send transactional mails through mailtrain


yashsaxena:
Is there a way to send transactional mails through mailtrain? But since transactional mails will not belong to any campaign & list will have a single email - how would that be triggered with mailtrain? I'm okay if there is a pure API way to fire this.

Additionally if this functionality needs to be built into mailtrain - how extensive is the effort?

goooseman:
We were looking for a solution for sending emails, we've tried mailtrain v2 and it is really excellent for us.

The only usecase we are missing is transactional mails.

I can add this functionality, but I need a little help with implementation details.

I've did a quick look at v2 source code and here is my first plan:

- [ ] Write an integration test
- [ ] Add `server/lib/template-sender`. It's a class, which has `templateId` in the constructor and a `send` async function, which takes an array of emails as a first argument and template params object as a second.
Short typing overview:
```
class TemplateSender {
  templateId: string;
  send(emails: string[], templateParams: Object) => Promise<void>
}
```
- [ ] Add POST `/templates/:templateId/send` route to `templates.js`.

@bures what do you think? If this feature is OK for you and for v2 release, I will be happy to make a PR with this feature.

bures:
Sounds like a good plan to me. I'll be very happy to merge this PR. Thanks.

As for the POST, if you want to use this as API (i.e. authenticated via access token), it would be better to put it to api.js (rather than in routes/templates.js). 

I assume that the number of emails send via a single call (i.e. the emails array in TemplateSender.send) will be relatively small, right? My guess, it won't really scale above tens or hundreds of addresses. Should the number be bigger, you would likely also need some asynchronous sending process and there we would get very close to campaigns functionality.

goooseman:
Yep, I want it to call with an access token.

In our usecase the number of emails will be always one.

I can make a restriction, so this API endpoint will not accept more than 100 addresses, so we will not have problems with asynchronous sending.

alexeychikk:
@bures Hi! I'm trying to implement what's discussed above and I have a question:
I retrieve template by id using `lib/models/templates` like this: `templates.get(templateId, (err, template) => { /*...*/ })`.  
Then I use `lib/mailer` to send the template like this:
```
mailer.sendMail(
   { 
     from: { address: settings.adminEmail },
     to: { address: email },
     subject: input.SUBJECT
   },
   {
     html: template.html, // using template.html from db
     text: template.text, // using template.text from db
     data: input.DATA
   },
[...]

PR #573, opened 2019-03-31
feat(api): Transactional mail rest api

goooseman:
In this PR we've added an API endpoint to send any email template with any template parameters to any email addresses.

This is really handy to use mailtrain as a transactional email sending system.


goooseman:
Resolves #352

bures:
Thank you. I had a quick look. Looks good. However, before accepting the PR, I would like to kindly ask you if you could add the following two things to make it in line with the rest of the API and general Mailtrain strategy:

1) Please use the context of the user (obtained via `req.context`) instead of `getAdminContext`, such that we can have granular permissions also for this API endpoint.

2) Please add an optional send configuration ID parameter to the API. Generally the `system` send configuration should be used only for delivering password reminders, etc. Having the ability to choose the send configuration makes it possible to differentiate sender domains, etc.

Thanks.

alexeychikk:
@bures Done. Please take a look.

bures:
Looks good. One more question and one more small request:

1) Do we use the instance of the TemplateSender class more than once? It makes sense for the CampaignSender because there it is created once and then send is invoked multiple times. I'm wondering whether to have TemplateSender as a class isn't a bit overkill. However, no big problem, especially if you think it makes sense to send through the object multiple times.

2) It would be good if the tag replacement could handle also fallbacks, such that we use the same syntax in templates for both transactional and mass sending. See https://github.com/Mailtrain-org/mailtrain/blob/3b20ac5ce710ddeb9ceef7a5afc2195b0a8fd399/server/lib/tools.js#L176

Thanks a lot.

alexeychikk:
@bures 
> 1. Do we use the instance of the TemplateSender class more than once?
No, but just in case I've refactored it a bit more for possible future cases. So now class instance is useful when you want to have some shared (default) options across different send calls.
> 2. I've changed `lib/tools.js` a bit so now TemplateSender uses the same format logic as CampaignSender.

bures:
Perfect! Thanks a lot.

patchthecode:
Is there any documentation on how this works?

PR #579, opened 2019-04-03
Dockerfile update

goooseman:
- `node:10-14-alpine `image is used (because `jessie` is oudated and can not install packages
- added package-lock.json to make sure the same dependencies' versions are used
- added an option to pass redisHost, mongoHost, mysqlHost to use in Kubernetes cluster

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,16 @@
-FROM node:10.14-jessie
+FROM node:10.14-alpine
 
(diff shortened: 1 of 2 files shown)


goooseman:
The new file is tested for the local set up (`docker-compose-local`) and for a production set up in kubernetes cluster

bures:
Good. Thanks.

PR #591, opened 2019-05-02
[DO NOT MERGE] Change REST API endpoints to also accept list name as a list identifier

jntesteves:
List CIDs are unpredictable, can only be created at runtime, and currently there isn't an API endpoint to GET them for REST clients. We are integrating with Mailtrain's REST API and this became a problem, especially for our automated tests that need predictability, but also for automatic configuration of our production systems that shouldn't require human intervention.

Hence, this change adds support for calling the REST APIs using the list name as an option in addition to the CID.

We decided for this very simple solution, instead of just fixing the list GET endpoint and making two requests per use of the API, because this is much faster and simpler to use, and a collision of a list name and a CID is highly unlikely to happen.

PS. Following this we'll send in more PRs, APIs for automating list creation, and predictable access tokens

### DIFF
--- a/lib/models/links.js
+++ b/lib/models/links.js
@@ -337,7 +337,7 @@ function getSubscriptionData(campaignCid, listCid, subscriptionCid, callback) {
             return callback(new Error(_('Campaign not found')));
         }
 
-        lists.getByCid(listCid, (err, list) => {
+        lists.getByCidOrName(listCid, (err, list) => {
             if (err) {
                 return callback(err);
             }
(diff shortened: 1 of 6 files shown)


jntesteves:
Update: we have discussed this change internally and we might come up with a better solution of setting List IDs in a more predictable way on list creation through the REST API (creating a new endpoint), since the List Name is not a unique key which makes this PR's implementation a little brittle.

I'll leave this open until we have a new implementation, since this works and is what we are using today, so it might be of interest to people.

PR #604, opened 2019-06-05
fix: added imagemagick to dockerfile;

goooseman:
imagmagick is required for resizing images inside Mosaico template
So I've added it to Dockerfile


bures:
Perfect. Thanks.

PR #610, opened 2019-06-12
Standardization of error handling

softionsrl:
Probably not the best way, but this should make the code more readable as a lot of it is just outputting error messages.
This allows for centralized error messages translation.


bures:
Thanks.

PR #618, opened 2019-06-19
allow attachments from api

### LINKED ISSUE #615: V2: Allow adding attachments to sendTemplateEmail API
Transactions emails will often require to send PDFs, etc as attachments(invoices, bills), etc. We could accept attachments as base64 encoded, and as urls, since multipart-encoding is not supported for API

heroic:
fixes #615


heroic:
@bures Can we merge this?

Issue #631, opened 2019-07-11
V2 bug: Mailtrain doesn't notice repeated emails with different casing


priethor:
When the same email addres is imported twice in the same list but with different letter casing, Mailtrain accepts them as different subscribers. However, when sending campaigns, all emails are converted to lower case, which results in the campaign being sent twice to the same address, but an error occurs when inserting the statistics into the DB, as the email address, already converted to lower case, already exists.

> Jul 09 17:23:13 xxx.xxx.xxx node[650]: ERR! Senders Sending message to 11:xxx@xxx.xxx failed with error: insert into `campaign_messages` (`campaign`, `list`, `response`, `response_id`, `send_configuration`, `status`, `subscription`, `updated`) values (113, 11, '250 Message queued as 16bd71df5c3000fdfc', '16bd71df5c3000fdfc', 7, 1, 1222, '2019-07-09 17:23:13.420') - ER_DUP_ENTRY: Duplicate entry '113-11-1222' for key 'cls'

Steps to reproduce:
- Import/create a subscriber with lower case email address.
- Import/create it again but with the first letter capitalized.
- Send a campaign to the list used.
- The email will arrive twice to the address but Mailtrain won't count the second one properly, leaving error traces in the logfile.

priethor:
Fixed by #632

bures:
#632 is not really a fix. However, there should be a fix coming in a few days.

bures:
Should be fixed now. Could you please try?

priethor:
Perfect, can close the issue now! Thanks!

PR #632, opened 2019-07-12
(#631) Bug correction

Charlo270398:
Hi, I fixed this bug. Now every mail is translated to lowercase before insert, import or update. Just consider changes into server/models/subscriptions.js file. Also I fixed another bugs, (adding lists into a campaign, re-subscribing by mail and edit a campaign namespace.


bures:
Thank you for the pull request. I'm afraid I cannot accept it, because it contains it does not handle `hash_email` and contains some garbage (e.g. your password to DB). Nevertheless, I took some important parts from it and fixed this bug myself. The fix will be available in a few days.

PR #637, opened 2019-07-22
Copy custom forms functionality

Charlo270398:
It would be useful to avoiding repetitive copy-paste, especially when a custom form must be translated in different languages. 
I just added a copy button at the end of the CUD forms page. The copied custom forms receive OLD_NAME+'_COPY' name.


bures:
Thanks. I'll be happy to accept the PR, however I would rather like this to work similar to templates. There is a "Clone from existing template" option when creating a template.

I guess, we should add another button to "Forms" list - something like "Copy from existing". This would lead to a create form where one selects an existing form and enters the name, description and namespace. Once this gets saved, it takes one to the edit screen. 

This has also the nice feature that no changes on the server are needed. 

One more thing - was the string "copy" automatically generated in locales/pt-BR/common.json or did you add it manually. The string should not be added there unless translated or marked with TODO. If it was generated there by the tool, then we have some bug in the tool. 

Thanks.

Charlo270398:
Hello,
I think I can make the changes in the same PR. I wrote the string "copy" by myself in locales/pt-BR/common.json, so there isn't a bug. I guess the correct word is 'Copia', like spanish word, but I'm going to mark it with TODO.

bures:
Perfect. Thank you. If you need any help or just want to discuss the feature before implementing it, let me know. I'll be happy to help.

Charlo270398:
Ok, d'akujem.

bures:
:-)

Charlo270398:
I just have finished, you can check it now.

Issue #538, opened 2019-02-17
v2 - Changing a campaign namespace results in a "Permission denied" error


priethor:
Steps to reproduce it the most simple use case:
1. Create a namespace child of the `Root` namespace, let's call it `N`.
2. Create a list belonging to namespace `N` (`L`)
3. Create a send configuration belonging to namespace `N` (`SC`)
4. Create a campaign belonging to namespace `N`, with list `L` and send configuration `SC`
5. Changing the namespace of `L` to `Root` results in a `Permission denied` error. No traces in log.

I have tried other ways flows for changing the namespace and they all result in the `Permission denied`

PR #641, opened 2019-07-23
Change campaign namespace BUG correction #538

Charlo270398:
I changed this line and it seems to work correctly. 
Also, I had problems adding multiple lists in a campaign (I just can add one). I fixed this too.


bures:
I believe this should be fixed now. I incorporated your changes in the fixes I made. Could you please check? Thanks.

Charlo270398:
Yes, it works now :+1:

PR #644, opened 2019-07-26
VERP test send BUG fixed #541

### LINKED ISSUE #541: v2 VERP - Wrong "Sender" header when sending template tests
Steps to reproduce:
1. Enable VERP
2. Edit a template content (not a campaign)
3. Send a test email with the template.

As VERP is enabled but we are not sending in a campaign context, the sender header in the resulting mail looks like this:

> Sender: [CAMPAIGN_ID].AWTsytRph.F-t196jPQ@xxx.yyy.zzz

The `[CAMPAIGN_ID]`  "merge tag" can't be resolved and the header has wrong characters for an email address, which makes email providers such as Gmail and Outlook reject the email.

Charlo270398:
I deleted campaignAddress field (a test is outside campaign context) and VERP behavior.

### DIFF
--- a/server/lib/campaign-sender.js
+++ b/server/lib/campaign-sender.js
@@ -85,8 +85,6 @@ class CampaignSender {
             return sendConfiguration[key];
         };
 
-        const campaignAddress = [campaign.cid, list.cid, subscriptionGrouped.cid].join('.');
-
         const mail = {
             from: {
                 name: getOverridable('from_name'),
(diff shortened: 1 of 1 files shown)


bures:
I was doing some bigger changes in the sending mechanism. Could you please recheck whether this PR still applies? I believe I fixed this in the process. Thanks.

Charlo270398:
It seems to be OK now.

PR #651, opened 2019-08-05
Clone from existing templates bug

### LINKED ISSUE #619: Cannot clone from an existing template
I'm using Mailtrain v2. When creating a template with "Clone from an existing template" option selected (from a Mosaico with predefined templates), I'm getting "There are errors in the form. Please fix them and submit again." The form is validated with all green tick and there's no error from the web console.

Charlo270398:
I have fixed #619 issue. Also I skipped code at _server/models/templates.js_ , if this last one is necessary it must be fixed. It returns an error (Cannot create property 'model' on string '{"mosaicoTemplate":1 ...}).


bures:
Thanks. I don't understand the changes in templates.js. Can you explain them?

Charlo270398:
If these lines are executed when you try to clone a template they return an error (Mosaico Templates). I just skipped them and it seems to clone correctly, but I don't know if these lines are really needed, so this should be fixed.

trantienthanh:
If I understand correctly, in server/models/templates.js
await files.copyAllTx(tx, context, 'template', 'file', entity.existingEntity, 'template', 'file', id);
is needed for copying files and data (in the database) from the existing templates?

I've been trying to fix this issue as well.

If we tick the checkbox "Clone from an existing template" without touching the template type or other template options (those options become hidden anyway), we will get this message "There are errors in the form. Please fix them and submit again". No other info can be seen in the console or in the log.

If we select the proper template type and template options, then tick the checkbox "Clone from an existing template", we will get this error message: "Cannot create property 'model' on string '{"mosaicoTemplate"

If I comment out this line:
//await _validateAndPreprocess(tx, entity);
in server/models/templates.js
then I get another error message: "insert into `templates` (`data`, `description`, `html`, `name`, `namespace`, `tag_language`, `text`, `type`) values ({"mosaicoFsTemplate":"versafix-1"}, '', '', 'Test Template 2', 1, 'simple', '', 'mosaicoWithFsTemplate') - ER_BAD_FIELD_ERROR: Unknown column 'mosaicoFsTemplate' in 'field list'"

I try to find the bit of code that handles all the SQL queries but it's quite hard as there's no documentation or anything like that for this project. Any suggestions would be greatly appreciated.

Charlo270398:
The error _Cannot create property 'model' on string '{"mosaicoFsTemplate":"versafix-1"}'_ is thrown by lines 26,27 and 28 in _server/lib/campaign-content.js_ , so this error is into the convertFileURLs function.

bures:
It should be fixed now. Thanks for help.

PR #654, opened 2019-08-06
Send configurations not working when VERP enabled bug

### LINKED ISSUE #653: V2: Send configurations not working when VERP enabled
Affected version: 2019-08-01-0745

When a send configuration has VERP enabled, campaigns can't be sent with that send configuration (test emails do work), failing with the following message:

> Aug 06 09:51:40 mailing.swiss4ward.com node[7957]: ERR! Senders Sending message to 13:1 failed with error: sendConfiguration is not defined.

@Charlo270398 already found the issue and will soon provide a PR.

Charlo270398:
Resolves #653


bures:
Thanks. I did some additional updates to it. Can you please test it now?

Charlo270398:
It seems to be ok :)

PR #661, opened 2019-08-09
Send configuration crash bug in campaigns CUD

### LINKED ISSUE #659: V2 Bug: users with a namespace N1 can see campaigns in namespace N1 that contain lists/SCs from N2
### Steps to reproduce:

1. Create a campaign and assign it to namespace N1.
2. Assign a list or send configuration of a different, non descendant, namespace N2.
3. Log in as a user with access to N1 but not N2 and go to campaigns.
4. The user can see and edit the campaign with settings outside his scope.

### Proposed solution:

If a campaign contains data a user shouldn't have access to, the user shouldn't see the campaign.

This should be improved by not allowing to save campaigns that contain send configurations or lists outside the campaigns scope (namespace subtree), as proposed in #562 and #563.

@Charlo270398 will provide a PR shortly.

Charlo270398:
Resolves #659

### DIFF
--- a/client/src/campaigns/CUD.js
+++ b/client/src/campaigns/CUD.js
@@ -97,7 +97,8 @@ export default class CUD extends Component {
         }
 
         this.state = {
-            sendConfiguration: null
+            sendConfiguration: null,
+            unallowedNamespace: false
         };
 
         this.nextListEntryId = 0;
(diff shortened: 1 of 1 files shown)


bures:
Thanks for the patch. It seems too complicated to me and possibly even a bit wrong. In my view, the right thing when catching the exception is to set sendConfiguration to null. If I remember correctly, the TableSelect that is there for the send configuration will realize that the send configuration is not permitted and will reset the send configuration in the form to null, which will eventually hide the whole send settings.

Thus, it should be just enough to catch the exception and set sendConfiguration to null. No need for all the fiddling with unallowedNamespace.

Charlo270398:
I thought that too, but there was a problem if I do this. The sendConfig selector shows this all the time: 

imagen
When a sendConfig is selected it disapears, so I don't know if is a real problem. I tried to skip the message adding that variable because sendConfig variable is already used to compare it with a null value, so I needed the extra variable to choose between show the message or not.

bures:
There was a bug in client/lib/table.js. I fixed that. I also fixed the campaigns/CUD the way we discussed above. There was also problem in  campaigns/Status, which I fixed too. Please have a look at a17ee3d9bfb5c9b6529092807ac3e26d3a0549f9 whether this solves the problem for you. If so, we can close this PR as the relevant fix is already part of the tree.

Charlo270398:
You commit seems to be perfect, it works find :ok_hand:

bures:
Perfect. Thanks.

PR #669, opened 2019-08-14
Added secure config parameter to use ldaps protocol

GuillaumeRemyCSI:
This PR adds the ability to use ldaps protocol, by using a "secure: true" parameter in ldap config.
Also improves parameters checking by using '===' instead of '=='.


bures:
Great. Thanks.

bures:
Actually, can you include the secure parameter to server/config/default.yaml, so that others know that the param exists? Thanks.

PR #671, opened 2019-08-14
Added new parameters for LDAP in docker-entrypoint.sh

GuillaumeRemyCSI:
Adds new parameters for docker-entrypoint.sh.
It allows to configure LDAP parameters dynamically when we launch Mailtrain using Docker


bures:
Perfect. Thanks.

PR #673, opened 2019-08-14
Create report BUG

### LINKED ISSUE #666: V2 Bug: create report interface crashes
Probably the CUD file has some errors.

Charlo270398:
Resolves #666


bures:
I'm afraid this is not the right fix. `user_fields` may be null until the user fields get loaded in `fetchUserFields`. If it throws an error because it is null, it is necessary to add corresponding if(s) that check if it is null.

Charlo270398:
Okey, I was wrong. Yes, that fixes the problem (iterate a null vector at line 81).

bures:
Perfect. Thanks.

PR #676, opened 2019-08-15
Fixes missing tx

heroic:
Use knex for queueing API send message instead of tx


bures:
I'm not sure what this is. In my view it contains a lot of unrelated files. If you need anything of this merged, please split it to distinct fixes/features that can be separately reviewed and approved. Thanks.

PR #679, opened 2019-08-15
Add mergeTags to sendMessage

heroic:


bures:
Thanks for the patch. Just to double check that I understand, can you describe what problem this solves?

heroic:
@bures sendMessage crashes when it doesn't receive the mergeTags.

bures:
OK. Thanks.

PR #680, opened 2019-08-16
Namespace filtering feature

### LINKED ISSUE #543: Feature request - Namespace filtering

Charlo270398:
Hi! I just developed this feature, allowing the namespace filtering. It resolves #543 issue. 
To active the filtering you must enable the _namespaceFilter_ option in the _default.yaml_ file. 

I added a link to namespace filter menu in the top bar. If a namespace is selected, this link shows the current namespace. In the namespace filter menu you can select a namespace, select it and return to the before page or turn off the namespace filtering.

When a namespace is selected you are only allowed to manage entities included in this namespace or his sub-namespaces. This filter is not active in all the pages because there are some exceptions exceptions where is useless, like in the lists show at campaign status page. 

I believe the implementation respects the permissions structure, because the filtering uses _where querys_ at the models functions, just adding extra filters to the current implementation.


bures:
Thanks for the patch. It's a good start. I would like to ask you for the following:

1) I would prefer local storage as opposed to cookie. It also means that the server does not have to get involved in setting the cookie, etc.

2) I would prefer passing the namespace filter in search query params of the REST call (as opposed to being part of the path). 

3) Please, instead of passing in the namespaceFilter to make the listDTAJAX methods as a parameter, pass it as `options` object that will contain the parameters from the search query. This way it will be more open for any future similar updates.

4) Please make the form to set the namespace as a modal (as opposed to being on a separate page). This way one can switch the namespace without having to leave unsaved data.

5) I admit, I didn't get the `getAllowedNamespaces` method. It looks way to complicate to me - also given the modifications in listTree. Also, using listTree in getAllowedNamespaces is an overkill. In this case, you don't have to care about permissions of children and you don't have to prune or sort the tree because the `dtHelpers.ajaxListWithPermissions` will take care of the permissions itself. I would leave listTree as is and just implement sime getAllowedNamespaces from scratch. It should just select all namespaces and then select all namespaces that are in the subtree.

There might be a few other smaller issues, but I would review them once this bigger batch is in place.

Thank you.

Charlo270398:
I agree with all the points, I'm going to work on it. 
I have a few questions:

1. Should I create an extra field in users table where indicate the current namespace filter? 
2. Totally agree with this. 
3. Totally agree with this too. 
4. I should do a dropdown list with all the allowed namespaces for the user? In this case, I think is very difficult to do a listTree dropdown. An option would be indicate the sub-namespaces like routes. For example (_Namepace1/Namespace1.1/Namespace1.1.2_). 
Another possible alternative would be reuse the current **Namespaces Filter** link, but showing a modal with a namespaces tree where select the filter namespace or disable it (similar to the separate page but into a modal).
5. I admit that is a complicate method too :sweat_smile:. The function tries to obtain all the sub-namespaces of a namespace (including itself) using a recursive function. I used listTree function to load the namespaces tree and apply this recursive function on the first node. I'm going remake this part as you indicate.

bures:
Perfect. Thanks.

Ad 1 - I wouldn't persist this on the server. Just per user in the local storage of the browser. 

Ad 4 - I'm not entirely sure I understand. However I was thinking to have the link on top right to open a modal with the namespace tree - essentially what is there now on the separate page.

Thanks. Let me know if you need any interim feedback.

Charlo270398:
Okay,  everything understood now.

Charlo270398:
I have the following problem introducing the **NamespaceSelect** into the modal at **Root** page:
_TypeError: "owner is null"_
The error comes from **TreeTableSelect** component (client/src/lib/form.js line 835), where this owner is needed. I don't know how to set that owner at Root page (like CUD pages, where NamespaceSelect is usually used).

bures:
Just create a wrapper component (with withForm) for it that wraps the NamespaceSelect in Form. Then include this wrapper component in root.

Charlo270398:
Ok, I forgot that I needed a form :man_facepalming: . It looks like this: 
imagen
I think there is a problem with .css because modal text is white, so I should fix this.

bures:
Perfect. Regarding the form, I would suggest the following:

- Use an "Enabled" checkbox (instead of the "Disable" button). I believe it will be more understandable.
- Make the checkbox and the tree list full width. There is an option for it `format="wide"`.

Here a suggestion for the buttons: "Apply", "Close". "Apply" would save the change and close the modal. "Close" would just close it without saving the change.

Charlo270398:
Maybe something like this? 

imagen

I still having the style problem (modal text is white, I guess the modal is not loading the correct .scss file).

client/src/root.js
```diff
@@ -35,7 +45,117 @@ if (mailtrainConfig.reportsEnabled) {
...
+    }
+
+    static propTypes = {
+        visible: PropTypes.bool.isRequired,
+        onHide: PropTypes.func.isRequired,
+    }
+
+    localValidateFormValues(state) {
+        const t = this.props.t;
+        if (!state.getIn(['namespace', 'value']) && state.getIn(['namespace_filter_enabled', 'value'])) {
```
bures: For fields that are not directly reflected in DB, I use camel case. I.e. namespaceFilterEnabled.

client/src/root.js
```diff
@@ -35,7 +45,117 @@ if (mailtrainConfig.reportsEnabled) {
...
+    @withAsyncErrorHandler
+    async loadTreeData() {
+
+            const response = await axios.get(getUrl('rest/namespaces-tree'));
+            const data = response.data;
+            for (const root of data) {
+                root.expanded = true;
+            }
+
+            if (this.props.entity && !this.isEditGlobal()) {
```
bures: This whole if is not needed here. `props.entity` is not defined.

client/src/root.js
```diff
@@ -35,7 +45,117 @@ if (mailtrainConfig.reportsEnabled) {
...
+
+        return (
+            <ModalDialog hidden={!this.props.visible} title={t('namespaceFilter')} onCloseAsync={() => this.hideModal()} buttons={[
+                { label: t('apply'), className: 'btn-primary', onClickAsync: ::this.setNamespaceFilterAsync },
+                { label: t('close'), className: 'btn-danger', onClickAsync: ::this.hideModal }
+            ]}>
+                <Form stateOwner={this}>
+                    <TreeTableSelect id="namespace" format="wide" label={t('namespace')} data={this.state.treeData}/>
+                    <hr/>
+                    <CheckBox id="namespace_filter_enabled" format="wide" text={t('enabled')}></CheckBox>
```
bures: Personally, I would put the checkbox above it and I would show the TreeTableSelect only when the checkbox is checked.

client/src/root.js
```diff
@@ -49,11 +169,16 @@ class Root extends Component {
 
         // The MainMenu component is defined here in order to avoid recreating menu structure on every change in the main menu
         // This is because Root component depends only on the language, thus it is redrawn (and the structure is recomputed) only when the language changes
+        @withComponentMixins([
+            withTranslation,
+            withForm
```
bures: Why does the MainMenu component has the `withForm`?

client/src/root.js
Charlo270398: Ok, that's a good point

client/src/root.js
Charlo270398: I was testing  this because I couldn't put the NamespaceTreeSelector, but I solved and I forgot to remove this.

Charlo270398:
Ok, I guess this point is done. It looks like this now: 

imagen
imagen

Finally I solved .css problem.

Charlo270398:
I'm having problems when a namespace is changed into a CUD/List page, I don't know how to refresh the tables according with the new namespace without refresh the entire page, loosing the introduced information.
The point is that I don't know how to connect the change in the local storage with the CUD page because both are in different page files but they share the same interface. I tried using Events but I can't change the CUD page state using this. Do you find a possible solution? There are two possible alternatives that I could do easily: 

1. Put namespace filter modal directly into the CUD pages.
2. Put a checkbox/button that applies the namespace filter into the CUD page.

bures:
My suggestion would be to store the namespaceFilterId in a context (on the root level). Then make the TableaSelect and Table to use the context is instructed so by a property.

bures:
I mean React context.

Charlo270398:
Ok, I'm going to try it.

Charlo270398:
a52820e this should fix points 2 and 3?

Charlo270398:
I'm having big troubles to implement the namespaceFilter context. I have realized that I can reproduce the re-render effect changing the language, so I have used _i18n.changeLanguage_ function for this, avoiding parameters (it maintains the current language). This should be fixed. 
Also I have fixed _getAllowedNamespaces_ function, now _listTree_ function is not modified.

Charlo270398:
I'm going to re-start this PR in #695

PR #681, opened 2019-08-16
LDAP edit users BUG fixed

### LINKED ISSUE #540: v2 LDAP - Existing users can't be updated
With LDAP enabled, users can be created and deleted, but editing them, even saving the edit form without changing any values, results in a `There are errors in the form. Please fix them and submit again.`

This is probably related to some fields (email, full name) being mandatory but hidden when LDAP is activated, even if the users already have values in those fields.

Charlo270398:
Resolves #540.


bures:
Thanks.

PR #687, opened 2019-08-21
LDAP - Users not created on the fly upon first login fixed #683 (good one) 

### LINKED ISSUE #539: v2 LDAP - Users not created on the fly upon first login
Quoting myself from #505:

> I've been testing the LDAP and can confirm that the login itself works fine with `passport-ldapauth`. However, I had to create the users internally in the user management page before they could login. I suspect this is not the desired use case and users should be created on first login, as Mailtrain accepts the ldap properties `newUserRole` and `newNamespaceId`, right?

When logging in with a wrong username & password pair, the form gives a `Invalid user or password` message, whereas using a valid user and password combination, an **permission denied error** appears.

Charlo270398:
Resolves #539

### DIFF
--- a/server/lib/passport.js
+++ b/server/lib/passport.js
@@ -187,9 +187,9 @@ if (LdapStrategy) {
 
             return {
                 id: user.id,
-                username: user.username,
+                username: profile[config.ldap.uidTag],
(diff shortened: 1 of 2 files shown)


bures:
Perfect. Thanks.

PR #695, opened 2019-08-30
Namespace Filter Feature

Charlo270398:


Charlo270398:
I used context this time, I think it solves the problem about i18n.changeLanguage() wrong use.

bures:
@Charlo270398 - sorry for being unresponsive in the last two months. I believe I should be more active now. Please, what is the status of this PR? Is it in the state I should take a look and possibly adjust it/integrate it in devel? Thanks.

priethor:
Hi @bures . I'm afraid Charlo is no longer part of our team and probably won't continue working on Mailtrain in the near future, as he might be busy with other projects. However, this PR should be ready for testing and, in the event of requiring more changes, another team member could continue working on it. Please let me know how to proceed, and you can adddress me in other PRs or issues where he is involved that are still open! 👍

Issue #374, opened 2018-02-02
Problem with images


jonathangilpin:
hi,

I am having trouble with images in a new install of mailtrain with mosaico and grapejs.

Summernote allows me to upload and insert an image OK.

However, mosaico and grapejs allow me to upload an image but then don't display it. They display a blank outline where the image would be but not the image itself.

Having done some searching I understand this is most likely an issue with ImageMagic but I am getting no errors from the application.

I have also done some testing e.g.:
 convert http://195.78.95.100:2000/mosaico/uploads/1/yoga.jpg -resize 100x100 /tmp/test.jpg
works fine.
convert is in the path in /usr/local/bin/
my policy.xml contains:

<policymap>
  
  
  
  
  
  
  
  
  
  <policy domain="coder" rights="none" pattern="EPHEMERAL" />

  <policy domain="coder" rights="none" pattern="MVG" />
  <policy domain="coder" rights="none" pattern="MSL" />
  
  
  
  
  
  
  
  
  
  
  
  
</policymap>

Is it possible to turn on some debugging to find out why the images are not showing up or what is going wrong? and has anyone got any suggestions on what I can try to get images working in mosaico? 

BTW, I have mosaico installed as a standalone application on the same server and that works OK.

bures:
Sorry. Didn't really encounter this problem. Any particular minimal set of steps to reproduce it? 

The logging level is controlled in the config file - field 'level' under '[log]'. However the chances are that there are no useful log messages produced at all. In that case, you can just simply insert log.info or similar to the js code of your instance to do the debugging.

thibwa:
@jonathangilpin any news ? 
I did the same ...

Fresh new env and now uploading image give me blank image:
![image](https://user-images.githubusercontent.com/6686308/39626116-f5d78cee-4fa0-11e8-9c24-97ef8129b30b.png)

But upload seems working: 
![image](https://user-images.githubusercontent.com/6686308/39626131-0115afc8-4fa1-11e8-961a-8bf9dc88ad46.png)

I searched on the folder and nothing ...

Thank you for your help

thibwa:
@jonathangilpin 
For your information, the issue in my case, was the URL configured in settings screen. By default, the URL is localhost:3000. Then I was accessing Mailtrain from the IP ... stupid issue ...

I hope it can help you.

Thanks 
Thibaut

thibwa:
Good to hear that ! 

See u
Thibaut

[...]

PR #715, opened 2019-10-02
Update README

fwolfst:
The sentences are a bit long, but it prevents default installations to have issues with image uploads (#374).


bures:
Thanks.

PR #722, opened 2019-10-10
Allow select segments on triggered campaigns

notz:
Don't know why this is disabled. Tested everything and it's working as expected (#403).


bures:
Thanks for noticing. It seems I implemented everything and I didn't realized that it actually works with segments. :-)

PR #723, opened 2019-10-10
Added subscription changed trigger

notz:
Is very useful if some subscription data is updated by API and you want to trigger after that because segments (filters - #722) have changed.


bures:
Makes sense. Thanks.

PR #732, opened 2019-10-16
Reduce docker image size

heroic:


martialblog:
I don't think it's a good idea to use an unofficial Docker image

bures:
I'm here with @martialblog . I would like to avoid using unofficial docker images as a base.

heroic:
@bures we can always replicate the image here. The difference is size is massive.

martialblog:
I think, the main issue here is not the base image. Most of the size comes from the node dependencies.

```
docker run -ti --rm --entrypoint bash mailtrain/mailtrain
bash-4.4# du -sh shared/node_modules/
174.2M	shared/node_modules/
bash-4.4# du -sh server/node_modules/
282.4M	server/node_modules/
bash-4.4# du -sh client/node_modules/
288.1M	client/node_modules/
```

Trimming some weight here would be an idea. Maybe using some webpack/uglify magic.

heroic:
@martialblog I've used modclean for mostly that reason. It removes a lot of cruft from node_modules. I am using it in production without any effects.

heroic:
js-yaml is a dependency of some other dependency, however, mailtrain uses it to parse yaml config. The current dependency that includes js-yaml is a devDependency. The Dockerfile in use, install all node_modules, not just production ones. So to reduce usage even further, I changed it to use only production modules. This causes a failure with a missing js-yaml. Hence added it to a production dependency, as a direct requirement, instead of being a transitive one.

martialblog:
I've experiment with these changes, with and without using modclean.

You can check it out here:
https://github.com/martialblog/mailtrain/commit/7d555f92e471864125651a0a2c0277bdc858133e

Result: 
```
 ~/github/mailtrain/ docker images | grep mailtrain
mailtrain/mailtrain              with-modclean     059e52beef33        459MB
mailtrain/mailtrain              without-modclean  ce0ecb67072a        466MB
mailtrain/mailtrain              latest            36dd9d0226bb        729MB
```

I've also tested the application with docker-compose and it works as expected. @bures what do you think? @heroic If you want you can adopt these changes and update your branch. Modclean doesn't make that much of a difference though.

heroic:
@martialblog we're already using with modclean in production, but if @bures wants to continue without it, we're okay.

bures:
Is this still relevant given the PR #792? If so, please make a new PR based on the current Dockerfile. At this point. Thanks a lot.

PR #740, opened 2019-10-21
Report triggered campaign messages & fix trigger of trigger

notz:
This fixes that triggered campaign e-mails are not reported in ui and so it's also possible to
create another triggered campaign on this first triggered one.

What i don't understand why their is a difference in handling of a triggered campaign and a normal one in sending.

Also i expected that a triggered campaign can only be send once to an user regardless which trigger fires the send.
But currently it's send for every trigger that is defined in a campaign.

### DIFF
--- a/server/lib/message-sender.js
+++ b/server/lib/message-sender.js
@@ -579,6 +579,16 @@ async function sendQueuedMessage(queuedMessage) {
     }
 
     if (messageType === MessageType.TRIGGERED) {
+        await knex.raw(knex('campaign_messages').insert({
+            campaign: campaign.id,
+            list: result.list.id,
+            subscription: result.subscriptionGrouped.id,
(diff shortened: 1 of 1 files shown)


bures:
I think that I never though about these different corner cases when I was porting the code from v1 to v2. So, any input (or even better specs or PR) are welcome.

bures:
I checked it. I think the correct behavior is that a triggered campaign should be delivered to a particular recipient only one regardless of how many triggers it has.

bures:
What I'm seeing in the code, that should be the implemented behavior - i.e. that a triggered campaign is delivered only once regardless of the number of triggers. Or do you experience any other behavior?

I'll merge your PR, but I'll remove the "ignore" in insert. In my view it should not be necessary there. Once I do it, I would like to ask you to test it. Thanks.

PR #751, opened 2019-10-26
Updated to allow JSON response from content url

ericuldall:


bures:
Thanks for the patch. Good idea. I guess, we need only some other way of checking whether this is JSON. I would like to avoid the guesswork by trying to parse the JSON. I think the best would be to check the content type of the response. If it returns "application/json", we should treat it as json. If it returns anything else, let's treat it as HTML for backward compatibility. Of course, it would be better if  in case of HTML it returned "text/html".

Would you please update it this way? I don't have a testing env for URL-type campaign set up. Alternatively, I can test it if you provide a test URL for the URL-type campaigns, I can update it myself. 

Thanks.

bures:

Issue #761, opened 2019-10-31
V2: Double sending


ericuldall:
I believe this application is double sending mail. I'm seeing database entries with 250 OK responses from the smtp server and status = 5.

ericuldall:
I'm seeing tons of this in my logs:
```
ERR! Senders Sending message to 1:2698 failed with error: Cannot read property 'id' of undefined.
```

ericuldall:
I beefed up my MySQL instance and it seems to be improving the double send issue. I'm still seeing the id errors, but not seeing any sent mail status 5 now. However, this is all indicative of some issue in the software. I'm going to triage this some more. I'll update any other findings here.

ericuldall:
Here is the line throwing the id error:

> verb TypeError: Cannot read property 'id' of undefined at MessageSender._sendMessage (/app/server/lib/message-sender.js:321:59)
```
list = this.listsById.get(listId);
```

bures:
Really strange. Is it happening deterministically? So that I can test it myself?

ericuldall:
I'm still trying to find a root cause. What I can tell is the list is missing from the listsById Map.

ericuldall:
I am having an issue with an email that is stuck on another campaign...and the list from the campaign that is stuck shows in the Map, but not the list from the campaign I'm looking at now.

ericuldall:
I'm adding extra logging to see at what point the list is not being added and perhaps that will turn something up.

bures:
Thanks. This part of Mailtrain is really complex. It's very likely there is some bug there.

ericuldall:
So I forced the old campaign into status 3 (the one that had the broken message), then I restarted the instance. Now my big campaign has the correct listById data in the map for the current campaign and seems to be running normally.

We definitely need to dig deeper here, I don't understand what went wrong. It seems to be related to my previous campaign being stuck in sending status but it was not able to send that last message.

ericuldall:
Perhaps a simple fix could be:
```
if (our list isn't in the map) {
  just grab it from mysql and add it anyway
}
```

bures:
No problem to do it, but unless we understand what is happening, I'm reluctant to add fixes that address only the symptoms. This could also be just a tip of an iceberg of some much worse problem.

ericuldall:
Yes, I agree. I'm going to keep digging on this for sure. I'm going to be gone tomorrow for one week, so probably have to dig more when I get back.

[...]

PR #763, opened 2019-10-31
Allow possibility of FAILED messages from content URL

ericuldall:


server/lib/message-sender.js
```diff
@@ -504,13 +516,21 @@ class MessageSender {
         try {
             result = await this._sendMessage({listId: campaignMessage.list, subscriptionId: campaignMessage.subscription});
         } catch (err) {
-            await knex('campaign_messages')
+            if (err.code === CampaignMessageErrorCode.NORETRY) {
+              await knex('campaign_messages')
+                .where({id: campaignMessage.id})
+                .update({
+                  status: CampaignMessageStatus.FAILED,
```
ericuldall: This will add status = -1 to mysql and I'm not really sure about the implications of that as of now. My hope is the current code will not try to send that message anymore. We should add some code that will allow us to present errors for failed messages in the UI as well as the ability to Retry Failed Messages.

ericuldall:
This addresses the issue detailed in https://github.com/Mailtrain-org/mailtrain/issues/761

ericuldall:
I just tested this on a sample campaign with an intentionally bad URL and it handles the errors okay. Marks the message as status = 6 (FAILED) and the campaign ended normally. This is just a start and allows me to look into my sql to see how many messages are failing and it keeps garbage out of the event loop.

shared/campaigns.js
```diff
@@ -53,7 +53,13 @@ const CampaignMessageStatus = {
     COMPLAINED: 4,
     SCHEDULED: 5,
 
-    MAX: 5
+    MAX: 5,
```
bures: Please set MAX to 6

server/lib/message-sender.js
bures: Indeed. It is necessary to go though the code and look for CampaignMessageStatus and figure out where the FAILED needs to be added.

shared/campaigns.js
ericuldall: Hi @bures ,

MAX wsa previously set to 5, what are the implications of changing it to 6? I'm not sure what the purpose of MAX is.

bures: MAX is used only to indicate the largest value of the enum. It's there if one needed to easily test that an option belongs to the enum.

PR #772, opened 2019-11-09
Fix rendering of access token in API docs

sedrubal:
...and remove trailing `&amp;`s.

#### Before:
Screenshot_20191110_003922

#### After:
Screenshot_20191110_003733


bures:
Thanks.

PR #773, opened 2019-11-09
Add Field `cid` to `/api/lists/:email`

sedrubal:
With this `cid` you are able to do further API calls like unsubscribing the user from all mailing lists.

Maybe it would be useful to add a separated API endpoint for that to streamline the calls.

#### Example response before:

```json
{
    "data": [
        {
            "id": 1,
            "name": "Test List"
        }
    ]
}
```

#### Example response after:

```json
{
    "data": [
        {
            "cid": "1-oXzj0Kw",
            "id": 1,
            "name": "Test List"
        }
    ]
}
````


bures:
Looks good. Thanks.

Issue #769, opened 2019-11-07
V2: Issue sending multiple campaigns at the same time


ericuldall:
When sending two or more campaigns at the same time my logs are flooded with the following error and sending makes no noticeable progress.

```
verb TypeError: Cannot read property 'id' of undefined
    at MessageSender._sendMessage (/app/server/lib/message-sender.js:335:59)
```

I'm assuming this may be due to some scoping issue somewhere, but not sure as of now. For my current usage I definitely need to be able to run campaigns in parallel otherwise my sends will never complete each day.

Anyone else experiencing this?

ericuldall:
Adding more context here since my line numbers will be off from the main development branch. Here is the exact line of code throwing the error:
```
const flds = this.listsFieldsGrouped.get(list.id);
```

ericuldall:
Found something curious...
```
if (subData.listId) {
            let listId;

            if (subData.subscriptionId) {
                listId = subData.listId;
                subscriptionGrouped = await subscriptions.getById(contextHelpers.getAdminContext(), listId, subData.subscriptionId);
            }
```

I'm wondering why `listId = subData.list` is inside of that if statement...? If for whatever reason subData.subscriptionId is not set, then there will be no list ID available for the call to listFieldsGrouped

ericuldall:
Reading more and it seems that both listId and subscriptionId are required, so it seems for some reason when sending multiple campaigns there is some issue with either the subscription id or multiple lists being set in the listsFieldsGrouped Map();

ericuldall:
After doing some more debugging, the issue seems to only occur when sending campaigns from multiple lists. I can send multiple campaigns from a single list ok.

ericuldall:
It seems the wrong values are getting passed to the `processCampaignMessages` function when running multiple campaigns. I've added some debugging and an see that function is getting called with a campaign Id of 6, but the messages array has messages from campaign Id of 5. Going to keep following this thread to see where the issue arises. 

@bures If you have any insight here that would be helpful. I'm slowly working my way to the solution.

ericuldall:
Could this be the issue in the master?
```
const subs = await knex('campaign_messages')
                .where({status: CampaignMessageStatus.SCHEDULED})
                .whereNotIn('hash_email', messagesInProcessing.map(x => x.hash_email))
                .limit(retrieveBatchSize);
```
Do we need to add the campaignId to this query?

ericuldall:
I'm going to try this:
```
[...]

PR #775, opened 2019-11-12
Fixes issue sending multiple campaigns

ericuldall:
Addresses issue https://github.com/Mailtrain-org/mailtrain/issues/769


bures:
Good catch. I didn't try it, but the patch looks logical to me. Thanks.

PR #780, opened 2019-11-13
V2: Return external ID (cid) in /api/lists/:email endpoint

Pistus:
Add *cid* to the list of fields returned per list for the */api/lists/:email* endpoint. This allows the response to be used for further api calls like updating a subscription, etc.

I kept *id* in the response for backwards compatibility.


Pistus:
Ha! Bummer - didn´t check PRs first and this is solved by https://github.com/Mailtrain-org/mailtrain/pull/773

PR #783, opened 2019-11-21
Refactor Docker entrypoint

martialblog:
Hello,

I've refactored the Docker entrypoint to be more in-line with regular deployments. 

- This provides a setup which is more consistens with other Docker images (using environment variables)
 - Provides the possibility to deploy configuration files with configuration management tool
- Adds a bit more verbosity with log output

I also added the list of Environment Variables into the README.

Cheers, Markus

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,17 +193,35 @@ These are the steps to start Mailtrain via docker-compose:
     docker-compose up
     ```
 
-    You can specify Mailtrain's URL bases via the `MAILTRAIN_SETTINGS` environment variable as follows. The `--withProxy` parameter is to be used when Mailtrain is put behind a reverse proxy.
-    ```
-    MAILTRAIN_SETTINGS="--trustedUrlBase https://mailtrain.example.com --sandboxUrlBase https://sbox.mailtrain.example.com --publicUrlBase https://lists.example.com --withProxy" docker-compose up
-    ```
-
 3. Open the trusted endpoint http://localhost:3000
 
 4. Authenticate as `admin`:`test`
 
 The instructions above use an automatically built Docker image on DockerHub (https://hub.docker.com/r/mailtrain/mailtrain). If you want to build the Docker image yourself (e.g. when doing development), use the `docker-compose-local.yml` located in the project's root directory.
 
+### Docker Environment Variables
+
+| Parameter        | Description |
+| ---------        | ----------- |
+| URL_BASE_TRUSTED | sets the trusted url of the instance (default: http://localhost:3000) |
+| URL_BASE_SANDBOX | sets the sandbox url of the instance (default: http://localhost:3003) |
+| URL_BASE_SANDBOX | sets the public url of the instance (default: http://localhost:3004)  |
+| WITH_PROXY       | use if Mailtrain is behind an http reverse proxy                      |
+| MONGO_HOST       | sets mongo host (default: mongo)                                      |
+| REDIS_HOST       | sets redis host (default: redis)                                      |
+| MYSQL_HOST       | sets mysql host (default: mysql)                                      |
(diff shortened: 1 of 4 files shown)


bures:
Looks good. Thank you. 

Have you tested it with LDAP? To me it seems that the `ldapUidTag` settings may cause problems. In the existing version, the `uidTag` entry is placed in the config only if the `ldapUidTag` is set. At this moment, however, it is always included, albeit empty.

martialblog:
I'll have a look at it and test it with an LDAP server

bures:
Thank you.

martialblog:
Hi, so I tested the config. Worked as expected.

If not present, the server/config/production.yaml gets initialized first with everything but the LDAP settings. Then this check kicks in:

```
    # Manage LDAP if enabled
    if [ "$WITH_LDAP" = "true" ]; then
        echo 'Info: LDAP enabled'
    cat >> server/config/production.yaml <<EOT
    ldap:
      enabled: true
      host: $LDAP_HOST
      port: $LDAP_PORT
      secure: $LDAP_SECURE
      bindUser: $LDAP_BIND_USER
      bindPasswort: $LDAP_BIND_PASS
      filter: $LDAP_FILTER
      baseDN: $LDAP_BASEDN
      uidTag: $LDAP_UIDTAG
EOT
    else
        echo 'Info: LDAP not enabled'
    cat >> server/config/production.yaml <<EOT
    ldap:
      enabled: false
EOT
```
If LDAP is disabled, application starts without having requiring futher settings.

bures:
Perfect. May I please ask you for one more thing? Since this changes how the docker containers are to be used and since some people may already rely on it, can we at least issue some warning from docker-entrypoint.sh when MAILTRAIN_SETTING variable is present? Also, it would be good to have an example how to invoke docker with the env. variables in the README.md. Thanks.

martialblog:
That's a good idea. You think just a warning is enough or should the entrypoint stop when it's present?

martialblog:
Soft Warning
```
if [ ! -z "$MAILTRAIN_SETTING" ]; then
    echo 'Warning: MAILTRAIN_SETTINGS is no longer supported. See README'
fi
```

Hard version:
```
if [ ! -z "$MAILTRAIN_SETTING" ]; then
    echo 'Error: MAILTRAIN_SETTINGS is no longer supported. See README'
    exit 1
fi
```

bures:
I think the hard version is better. I believe that when something fails, 
it's better if it fails hard. Easier to debug. Thanks.
T.

------ Original Message ------

martialblog:
Alright. I just pushed the changes.

bures:
Great. Thanks.

gerhardsletten:
@martialblog After this change it looks like that to pass env-vars into the container you will need to create a `docker-compose.override.yml` that extend the mailtrains config enable own override of env-vars for either a `.env` file or as env-vars prepended to the docker-compose up command? 

```
version: '3'

services:
  mysql:
    command: mysqld --sql_mode=""
  mailtrain:
    environment:
    - URL_BASE_TRUSTED
    - URL_BASE_SANDBOX
    - URL_BASE_PUBLIC
    - WWW_PROXY
```
[Ref docker-compose doc about env-vars here](https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers)

Before this change you could just add all settings in the same into the docker-compose up, but if you just replace MAILTRAIN_SETTINGS with single options they will not be picked up in docker-entrypoint.sh

```
MAILTRAIN_SETTINGS="--trustedUrlBase https://mailtrain.example.com --sandboxUrlBase https://sbox.mailtrain.example.com --publicUrlBase https://lists.example.com --withProxy" docker-compose up
```
Maybe this should be explained in the README.md?

PR #789, opened 2019-12-03
Fix trigger campaign id

### LINKED ISSUE #752: Triggers don't work in V2
https://github.com/Mailtrain-org/mailtrain/blob/3a2d1512ab22472ba92c8adff2a09a918a4cbba0/server/services/triggers.js#L68

I think it's because this should be trigger.source_campaign instead of `campaign.id`. `campaign.id` if the triggered campaign the trigger belongs to, and has yet to sent.

notz:
fixes #752


bures:
Yep. Good catch. Thanks.

PR #792, opened 2019-12-05
Reduce Docker image size

martialblog:
Hi, 

I refactored some stuff to reduce the image size a bit.

> Before; 729MB
> After: 548MB
> Reduction: 181MB

Since the client node_modules are only needed during the build phase, the directory can be deleted afterwards. However, the server/app-builder.js serves some JS from the client directory, therefore I added copy-webpack-plugin to copy these to the client/dist.

```
 /static-npm/popper.min.js 
 /static-npm/coreui.min.js 
 /static-npm/bootstrap.min.js 
 /static-npm/jquery.min.js
 /static-npm/fontawesome/fa-solid-900.woff2
```

Let me know if you want some changes to the PR.

This PR makes https://github.com/Mailtrain-org/mailtrain/pull/732 obsolete

Cheers, Markus


martialblog:
I just pushed another tiny change to the branch, to futher reduce the size.

We're now at:

> Before: 729MB
> After: 472MB
> Reduction 257MB

bures:
Thanks.

PR #803, opened 2019-12-12
Override config for docker-compose

gerhardsletten:
Adjustments for docker-entrypoint.sh to be able to add own server/config/production.yaml when building an own image of mailtrain, and some notes about passing env-vars into docker-images after #783


bures:
There seems to be a merge conflict. If still relevant, could you please update the PR so that it can be merged? Thank you.

gerhardsletten:
@bures Hi, it is updated now

bures:
Please check whether this is still in line with your requirements. I had to merge it with other code. Please check also the README.md whether it is factually correct. I rewrote the part explaining the settings. It couldn't understand it before. Thanks.

Issue #748, opened 2019-10-24
V2: List import sending TIMESTAMP to DATETIME column


ericuldall:
It seems we are unable to import date columns into our lists. I checked the mysql schema for Date types and it is using `DATETIME`, but the import is sending dates formatted like `1970-01-01T00:00:00.000Z` a timestamp.

I'm happy to work on a PR for this if you can point me to the relevant files.

ericuldall:
Error from import log:

```
ER_TRUNCATED_WRONG_VALUE: Incorrect datetime value: '1970-01-01T00:00:00.000Z' for column 'custom_dob_2ppihxu7f' at row 1
```

gerhardsletten:
Have the same problem, when I inspect the data-flow I see that field-value passed to [parsePostValue-func here](https://github.com/Mailtrain-org/mailtrain/blob/development/server/models/fields.js#L246) contains a setting-field that is a stringified object:

```
RowDataPacket {
  id: 3,
  name: 'Last Order Date',
  type: 'date',
  help: null,
  key: 'MERGE_LAST_ORDER_DATE',
  column: 'custom_last_order_date_3hzxci7lv',
  settings: '{"dateFormat":"eur"}', <-- here
  group: null,
  default_value: null,
  order_list: 3,
  order_subscribe: 3,
  order_manage: 3
}
```
And this will give `parseDate` and undefined first param:

```
parseDate(field.settings.dateFormat, value)
```
Have not looked into how data is stored in the database, but should this `settings` property be a string or should it be parsed to an object? 

The other issue with custom date-fields is that when converting a js Date object to a string and this data is posted to the ajax-route, the date-time is not a valid mysql-date. [Read more here](https://stackoverflow.com/questions/44304777/er-truncated-wrong-value-incorrect-datetime-value)

But a quick-fix for this to change the sql_mode in the docker-file: (not sure if this safe though..)

```
version: '3'

services:
  mysql:
    image: mariadb:10.4
    environment:
    - MYSQL_ROOT_PASSWORD=mailtrain
    - MYSQL_DATABASE=mailtrain
    - MYSQL_USER=mailtrain
    - MYSQL_PASSWORD=mailtrain
    volumes:
    - mysql-data:/var/lib/mysql
    command: mysqld --sql_mode="" <-- Add this
```

ericuldall:
[...]

PR #804, opened 2019-12-12
Fix import of date and general handle of mysql-dates in docker

gerhardsletten:
Need to parse the fields settings to make the date-format available in parsing, and set sql_mode in docker to make dates work even if they are not strict mysql-dates

### DIFF
--- a/docker-compose-local.yml
+++ b/docker-compose-local.yml
@@ -10,6 +10,7 @@ services:
     - MYSQL_PASSWORD=mailtrain
     volumes:
     - mysql-data:/var/lib/mysql
+    command: mysqld --sql_mode=""
 
   redis:
     image: redis:5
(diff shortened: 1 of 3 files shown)


bures:
Thank you. Why is the sql-mode overridden in the docker?

gerhardsletten:
@bures The background is that custom date-fields dont work, at least for european dates, which I think is  think has to do with the way dates are passed to the db. When they are stored from import of directly from an input not in the strict format which mysql docker image is configured with.

The background is that we have a list where we have a custom field with the customers last shopping date, and then we filter out recent customer from email campaigns, and it was in the csv-import I discovered that this was not working in the v2 branch.

There is also a discussion here  #748 and a link to a [SO tread here](https://stackoverflow.com/questions/44304777/er-truncated-wrong-value-incorrect-datetime-value)

bures:
Please see my comment in #748.

bures:

PR #813, opened 2019-12-19
Handle failures from AWS SES

gerhardsletten:
If a message fail to be delivered by AWS SES check if retryable is set to false to abort new retries. Else mailtrain will continue to try to deliver it until retention level is reach for campaign (default to 1 day)

### DIFF
--- a/server/lib/message-sender.js
+++ b/server/lib/message-sender.js
@@ -516,7 +516,10 @@ class MessageSender {
         try {
             result = await this._sendMessage({listId: campaignMessage.list, subscriptionId: campaignMessage.subscription});
         } catch (err) {
-            if (err.campaignMessageErrorType === CampaignMessageErrorType.PERMANENT) {
+            if (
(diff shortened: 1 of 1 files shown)


bures:
Thank you. Where does the `err.retryable` come from?

gerhardsletten:
@bures Its from the aws-sdk-js package [see the error-definition here](https://github.com/aws/aws-sdk-js/blob/master/lib/error.d.ts#L16)

bures:
Understood. What I'm not happy about is that an AWS specific detail leaks to code which should be independent of a specific mail service. This should be abstracted somewhere around server/lib/message-sender.js:451, where we already have some mail service specific code. 

I believe the right solution is to catch errors coming from server/lib/message-sender.js:451 and set the `campaignMessageErrorType` to `CampaignMessageErrorType.PERMANENT`. In the comment, this should be marked as AWS specific, similar to as it is in server/lib/message-sender.js:456-459

Do you think you can do this? Thanks.

bures:

PR #827, opened 2020-01-07
Fix missing tx refrence, https://github.com/Mailtrain-org/mailtrain/i…

Mubby:
Fix missing tx reference when sending transactional email through API

### DIFF
--- a/server/lib/message-sender.js
+++ b/server/lib/message-sender.js
@@ -682,7 +682,7 @@ async function queueCampaignMessageTx(tx, sendConfigurationId, listId, subscript
     });
 }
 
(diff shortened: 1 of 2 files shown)


bures:
Thanks for reporting. Actually, the right solution is in my view to change message-sender.js:698 to `await knex('queue')...` instead of adding the transaction there.

Mubby:
Given that the sendAsTransactionalEmail can take an array of emails to send out I feel like that is the best place to open the transaction. If any of those emails in the array fail then you could end up in state where half your emails had been committed for processing and some not making it difficult to recover from without sending duplicate emails.

bures:
Makes sense. I merged it and did some improvements to the code. Can you test it now, please? Reopen the issue if there are any problems. Thanks.

PR #838, opened 2020-01-25
fix endpoint /api/blacklist/delete

andresmrm:
Avoids error "Cannot read property 'user' of undefined" when using /api/blacklist/delete endpoint.


bures:
Thanks.

PR #842, opened 2020-01-31
Add API enpoints to create, delete and get lists.

andresmrm:
Adds:
- GET /lists-by-namespace/:namespaceId
- POST /list
- DETELE /list/:listCid

These are my first changes to a node backend, so please review (specially to check for security breaches) and feel free to request changes. I can also try to split this commit in one per endpoint if you prefer.

The paths of the current API endpoints seems a bit messed up... But I didn't changed them to avoid breaking changes. Only avoided path collision with the new endpoints.

Since now we have the REST API used by the admin client, should we eventually remove this old API and add token access to the REST API? Maintaining two APIs to do the same things seems strange.


renewooller:
Is there a reason this hasn't been merged?

bures:
Thanks. Sorry for the delay. The only reason is the lack of time.

Issue #505, opened 2018-12-25
Mailtrain v2 Beta Available - help needed with testing


bures:
Dear all,

it's my pleasure to announce general availability of Mailtrain v2 beta. You can find it in the `development` branch. The repo contains installation scripts for CentOS 7 and Ubuntu 18.04 LTS. There is also a Docker image with a docker-compose script provided. All info how to install it is in README.md.

This version should have all features of Mailtrain v1 with a bunch of new things added. Notably those are:
- multiple users and organisational namespaces with fine-grained user permissions (the mapping of user permissions to roles is defined in config file)
- various new features in lists (e.g. list export)
- various new features in templates (e.g. support for MJML, management of files, isolation of template editors to prevent XSS attacks)
- various new features in campaigns (e.g. ability to send to multiple lists with prevention of double delivery)
- multiple send configurations
- ability to localize the UI per user (as opposed to per installation as it was in v1)
- changes in the architecture (await/async on server side, ReactJS-based single-page application on the client side)
- builtin ZoneMTA, which is used as a default delivery mechanism (this requires close-to-zero configuration of the mailer)

Note however that there are several features which I didn't have a chance and possibility to test. In particular:
- LDAP
- delivery via AWS SES
- delivery via Postfix
- delivery via Sparkpost
- delivery via Sendgrid
- delivery via Mailgun
- bounce handling via VERP
- most of the API calls

I would like to ask you for help with testing these features above. If it turns out that there is no one to test the feature, I will assume that that feature is not need and will exclude it in stable release of Mailtrain v2. 

If you file an issue, please mark it with `v2.0` milestone.

Thanks. T.

hoducha:
I got an error when installing Mailtrain with `docker-compose`. Here is the full log: https://pastebin.com/AYMkBNCu

sweetppro:
Is it possible to install manually on a single domain?
I am running a Centos 7 server with Plesk. And I want to install this in a subdomain(as I have already with v1)

I see that we need 3 endpoints now though. Can I host all the endpoints in subdirectories, rather than seperate subdomains?

If that’s possible, then I’m totally in for some small delivery tests with Mailgun, as I have this all workin well for mailtrain v1.

bures:
@sweetppro - How strong is this requirement (of hosting on single endpoint)?

[...]

PR #844, opened 2020-02-02
Set more config options from docker-entrypoint.sh

taye:
This PR makes the docker entry point use the following new environment variables to set the production config:

| Parameter        | Description |
| ---------        | ----------- |
| PORT_TRUSTED     | sets the trusted port of the instance (default: 3000)                 |
| PORT_SANDBOX     | sets the sandbox port of the instance (default: 3003)                 |
| PORT_PUBLIC      | sets the public port of the instance (default: 3004)                  |
| WWW_HOST         | sets the address that the server binds to (default: 0.0.0.0)          |
| WWW_SECRET       | sets the secret for the express session (default: `$(pwgen -1)`)      |
| WITH_REDIS       | enables or disables redis (default: true)                             |
| WITH_ZONE_MTA    | enables or disables builtin Zone-MTA (default: true)                  |

Edit:

~~This changes the expected values of the `URL_BASE_*` variables so that they should exlude the ports which are now gotten from the `PORT_*` variables~~

Re: https://github.com/Mailtrain-org/mailtrain/issues/505#issuecomment-578184150

:heart:

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,9 +213,9 @@ variables (e.g. `URL_BASE_TRUSTED=https://mailtrain.domain.com (and more env-var
 | PORT_TRUSTED     | sets the trusted port of the instance (default: 3000)                 |
 | PORT_SANDBOX     | sets the sandbox port of the instance (default: 3003)                 |
 | PORT_PUBLIC      | sets the public port of the instance (default: 3004)                  |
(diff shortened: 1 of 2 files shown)


bures:
Looks good. Thanks.

PR #853, opened 2020-02-17
Update subscription widget docs

seesemichaelj:
The existing docs for the subscription widget are either vague (using `domain` instead of pointing out you should use the untrusted endpoint [which should be obvious]) or incorrect (missing `/static` when getting `widget.js`)


bures:
Thanks.

PR #859, opened 2020-03-03
fix for 'shortid not defined'

alangecker:
Fixes following error:
```
 ERR! Senders Sending message to 4:1 failed with error: shortid is not defined. Dropping the message.
 verb ReferenceError: shortid is not defined
     at html.replace (/home/mailtrain2/mailtrain/server/lib/message-sender.js:236:29)
     at MessageSender._getMessage (/home/mailtrain2/mailtrain/server/lib/message-sender.js:235:25)
     at MessageSender._sendMessage (/home/mailtrain2/mailtrain/server/lib/message-sender.js:345:34)
```


bures:
Thanks.

PR #889, opened 2020-04-30
French locale

rycks:
Hello mailtrain, here is a french translation for v2/beta ... some updates will become but this is a first step !


bures:
Thanks.

PR #900, opened 2020-05-25
Translate PT-BR. 

geisimarrech87:
Automated translations.


geisimarrech87:
Change with compiled file

Issue #716, opened 2019-10-03
V2: Campaign Statistics not updating


trantienthanh:
I'm running Mailtrain v2 with latest commit cbf2a6e39d3e40643c5e56f5867b7190079076ab.
I'm not quite sure the issue started from which commit but all my campaign statistics after August remain 0 (0%)
Blacklisted 0
Bounced 0 (0%)
Complaints 0 (0%)
Unsubscribed 0 (0%)
Opened 0 (0%)
Clicked 0 (0%)

No issue with sending emails. Success and bounced emails are all shown up in the log.
Any suggestion would be really appreciated.

trantienthanh:
Bounced status still get updated correctly in the Subscribers list.

trantienthanh:
verb HTTP Error: Not Found
verb HTTP     at Object.getTrackingSettingsByCidTx (/opt/mailtrain/server/models/campaigns.js:336:15)
verb HTTP  { Error: Not Found
verb HTTP     at Object.getTrackingSettingsByCidTx (/opt/mailtrain/server/models/campaigns.js:336:15)
verb HTTP   stack:
verb HTTP    'Error: Not Found\n    at Object.getTrackingSettingsByCidTx (/opt/mailtrain/server/models/campaigns.js:336:15)',
verb HTTP   type: 'NotFoundError',
verb HTTP   data: undefined,
verb HTTP   status: 404 }
Error [ERR_HTTP_HEADERS_SENT]: /opt/mailtrain/server/views/layout.hbs: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:470:11)
    at ServerResponse.header (/opt/mailtrain/server/node_modules/express/lib/response.js:771:10)
    at ServerResponse.send (/opt/mailtrain/server/node_modules/express/lib/response.js:170:12)
    at done (/opt/mailtrain/server/node_modules/express/lib/response.js:1008:10)
    at /opt/mailtrain/server/node_modules/hbs/lib/hbs.js:106:11
    at Object.done (/opt/mailtrain/server/node_modules/hbs/lib/async.js:74:20)
    at /opt/mailtrain/server/node_modules/hbs/lib/hbs.js:101:20
    at /opt/mailtrain/server/node_modules/hbs/lib/hbs.js:56:11
    at Object.done (/opt/mailtrain/server/node_modules/hbs/lib/async.js:74:20)
[...]

PR #904, opened 2020-06-11
small translation improvement for de-DE

wschopohl:


wschopohl:
for me and some more people (issue #716) bounces are not showing up in the statistics via webhooks (aws in my case). Second commit seems to fix this and was inspired by trantienthanh

bures:
Thanks.

PR #907, opened 2020-07-01
Docker Development Container

wschopohl:
Added slightly tweaked docker and docker-compose files, to easily get a local development server started. Also added a little section to the README.md file, to briefly explain what to do. I thought this might make it easier for other developers to start contributing.


bures:
Looks good. Would it be possible to exclude the json-lock files from the PR? I don't think there was any change in the files that would require change in the modules. Thanks.

wschopohl:
No problem! Wasn't sure how to deal with the lock files.. Should be gone now!

bures:
Perfect. Thanks.

PR #909, opened 2020-07-13
Use proper escaping for variable column names

andris9:
- Fixes issue with invalid column names

### DIFF
--- a/lib/models/campaigns.js
+++ b/lib/models/campaigns.js
@@ -60,10 +60,10 @@ module.exports.statsClickedSubscribersByColumn = (campaign, linkId, request, col
             return callback(err);
         }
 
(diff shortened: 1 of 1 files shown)


lib/models/campaigns.js
```diff
@@ -60,10 +60,10 @@ module.exports.statsClickedSubscribersByColumn = (campaign, linkId, request, col
             return callback(err);
         }
 
-        let query_template = 'SELECT %s AS data, COUNT(*) AS cnt FROM `subscription__%d` JOIN `campaign_tracker__%d` ON `campaign_tracker__%d`.`list`=%d AND `campaign_tracker__%d`.`subscriber`=`subscription__%d`.`id` AND `campaign_tracker__%d`.`link`=%d  GROUP BY `%s` ORDER BY COUNT(`%s`) DESC, `%s`';
+        let query_template = 'SELECT ?? AS data, COUNT(*) AS cnt FROM `subscription__%d` JOIN `campaign_tracker__%d` ON `campaign_tracker__%d`.`list`=%d AND `campaign_tracker__%d`.`subscriber`=`subscription__%d`.`id` AND `campaign_tracker__%d`.`link`=%d  GROUP BY ?? ORDER BY COUNT(??) DESC, ??';
         let query = util.format(query_template, column, campaign.list, campaign.id, campaign.id, campaign.list, campaign.id, campaign.list, campaign.id, linkId, column, column, column);
```
bures, line 64: Are you sure it works. Shouldn't the "column" parameters be removed from the format function?

PR #922, opened 2020-07-28
Fix plain text generation

### LINKED ISSUE #882: v2 plain-text doesn't get send
Hi, 
we use Mailtrain 2 beta, Build 2019-01-12-1254

As you can see in the pic, we add plain-text to this Template content (plain text). In the sent mails, we cannot find any of these. Is it a known bug?

mailtrain-v2beta-template-plain-text2020-04-16 22-14-58

Treora:
Due to one inversion too many, the email’s plain text version was overridden with version generated from the html; while if there was no plain text, none would be generated.

This might also fix #882, if its author meant that *their* plain text does not get sent.


bures:
Looks good. Thanks.

PR #926, opened 2020-07-29
Localization improvements

Britaliope:
### General
mailtrain version: v2

### Description

Localize strings that were not localized correctly


bures:
Thanks. Did you do this manually or using the script located in locales? I'm surprised that it didn't modify the other languages as well.

Britaliope:
> Thanks. Did you do this manually or using the script located in locales? I'm surprised that it didn't modify the other languages as well.

No, I did this manually as i didn't found documentation about localization

bures:
There is a script for this `locales/extract.js`. Anyway, before using it, the script has to be updated to reflect all the languages currently present. If you need to update the localization strings, please use the script. I'll close this PR now as it might cause problems with other languages.

Issue #925, opened 2020-07-28
V2 API Improvements


c-seeger:
Hello,

first of thanks for this awesome project it works like charm! Great work.

Since i needed a go client for your API i build one [mailtrain-go-api](https://github.com/Virtomize/mailtrain-go-api), feel free to use and share it if you want to.

While Implementing i found out a few issues when it comes to the API:

- The documentation sometimes refrences different path in the title and the curl example
  - e.g: `DELETE /api/lists/:listId – Delete List` and `curl -XDELETE 'https://mailtrain.virtomize.com/api/list/B16uVTdW?access_token=4798aa8a8ff3485e83f27c06edd8c74491b7af5f'` (see lists vs list)
- The usage of listID and listCID is not consistent as sometimes the responses give back the listID and the requests use listCID which can be a bit confusing sometimes
- `POST /api/blacklist/delete – Delete email from blacklist` is a Delete so when implementing strict REST this should use http DELETE Method resulting in a change of the path to `/api/blacklist/delete/:email`? or maybe using delete as an action verb `api/blacklist/:email/delete` since email is has no ID using the email itself in the url is a bit uncommon - so this one is just an idea to think about if that may or may not make sence
- same as above for `POST /api/delete/:listId – Delete subscription`
- the fieldnames are uppercased in the documentation but when using against the API lowercase version properly works maybe this could also be changed in the documentation
- the API documentation lacks of return types, would be awesome to have them there so no reverse engineering is needed :P

HTH to make this great project even better!

forgeOneHQ:
We are going to start with the development and testing of mailtrain v3 in the next weeks.

You are welcome to help us with the testing as soon as the first release candidate is available.

Issue #778, opened 2019-11-13
/api/lists/:email endpoint should return cid instead of id


Pistus:
All api calls are based on a lists `cid` and not the `id`. However the /api/lists/:email endpoint which returns a list of the email lists an email is on returns the id value. This makes it impossible to use the list to for example verify that an email is on a list before updating the subscription.

forgeOneHQ:
We are going to start with the development and testing of mailtrain v3 in the next weeks.

You are welcome to help us with the testing as soon as the first release candidate is available.

PR #935, opened 2020-08-20
API Improvements (v2)

### LINKED ISSUE #903: GET /lists/:email should return unsubscribed
Hello,
I want to know through API if an email has unsubscribed but it seems impossible for now.
Could we add this feature ?
Sincerely.

joker-x:
- Fix #903 #778 
- Working on #925

### DIFF
--- a/client/src/account/API.js
+++ b/client/src/account/API.js
@@ -86,16 +86,6 @@ export default class API extends Component {
                     </div>
                 </div>
 
-                <!--<div className="card mb-3">
-                    <div className="card-header">
-                        <b>POST /api/subscribe/:listId – {t('addSubscription')}</b>
-                    </div>
-                    <div className="card-body">
-                        <p className="card-text">
-                            {t('thisApiCallEitherInsertsANewSubscription')}
-                        </p>
-                    </div>
-                </div>-->
                 <h4>POST /api/subscribe/:listCid – {t('addSubscription')}</h4>
 
                 <p>
--- a/client/src/account/API.js
+++ b/client/src/account/API.js
@@ -86,7 +86,7 @@ export default class API extends Component {
                     </div>
                 </div>
 
-                <div className="card mb-3">
+                <!--<div className="card mb-3">
                     <div className="card-header">
                         <b>POST /api/subscribe/:listId – {t('addSubscription')}</b>
                     </div>
@@ -95,8 +95,8 @@ export default class API extends Component {
                             {t('thisApiCallEitherInsertsANewSubscription')}
                         </p>
                     </div>
-                </div>
-                <h4>POST /api/subscribe/:listId – {t('addSubscription')}</h4>
+                </div>-->
+                <h4>POST /api/subscribe/:listCid – {t('addSubscription')}</h4>
 
                 <p>
                     {t('thisApiCallEitherInsertsANewSubscription')}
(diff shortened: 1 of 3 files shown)


bures:
Sounds good to me. There will be probably some changes once the implementation starts, but overall looks good. Eventually I would like to end up with removing api v1 and keeping api v2, so I'm still thinking how should we name the endpoints.  Since we are still in beta and your changes in the PR break the v1 API anyway (at least as far as I understood them), maybe we can just require a parameter in the reqeust header that specifies the version of the API. Default will be 2. Then we can keep the endpoints as `api/....`. What do you think? Thanks.

bures:
Or even better, we have no default for the version. That means that if the version is not specified, and error will be throws. This would mitigate any problems where the endpoint has the same URL but does something slightly different than before.

joker-x:
This PR don't break the v1 API, only add new fields (cid and status) in response of /list/:email endpoint. Also return cid instead of id in /unsubscribe/:listCid and /delete/:listCid endpoint, but with the user id can't be call to none of other api endpoint. The input params don't changed, only fixed in the documentation.

I think that insert version of the API in a query param will make difficult implement a new version from scratch. I would like all endpoints of v1 serve in route /api/v1/... and endpoints of v2 in /api/v2/..., but this break routes for v1. This is the reason for use /apiv2/... Maybe can use /api/... for v1 and /api/v2/... for v2

joker-x:
I think that we must create documentation page of new api (including params and responses) after begin implementation to minimize breaking changes.

joker-x:
The main changes in draft of version 2 with respect to version 1, in addition to supporting the new entities that did not exist will be:

- Use verbs http as REST standard: GET for listing, POST for create, PUT for update and DELETE for delete.
- Use plurals in entities always, for simplicity.
- All entities that only make sense inside a list put into /apiv2/lists/:listCid...
- Unify /api/list/:email and /api/lists-by-namespace/:namespaceid into /apiv2/lists because are filter params
- Unify /api/subscribe/:listCid and /api/unsubscribe/:listCid into /apiv2/lists/:listCid/subscriptions because both are updates of its table subscription

bures:
Is this PR really non breaking? The changes in api.js seemed to me that they change the interface/semantics. But I could be wrong.

As for the proposal, I'm not concerned much about breaking the original API by requiring an extra header. Alternatively, for some time, we can assume that no header with version selector means v1. The reason why I'm not very much a fan of `v2` in the path is that I would like to start with a clean slate and not to keep in the path the heritage of originally having two APIs. 

We can easily have two different API routers (for v1 and v2) and just decide between them based on the header parameter which will specify the version (and default to v1 for the time being). 

I like the idea with the V2 API page. Just as I said, I would like to go with the paths /api/xxxx and specify the version in the header. It will also allow us to accommodate breaking changes in the API later on.

Thanks.

joker-x:
> Is this PR really non breaking? The changes in api.js seemed to me that they change the interface/semantics. But I could be wrong.
> 

For practical purposes for the cases of unsubscribe and delete it is irrelevant that they return the cid instead of the id because the subscribers are identified by the email, no other endpoint receives the id or the cid. I have changed them for consistency since subscribe returned the cid.

Changes in /list/:email do matter: it was impossible to show which lists you were subscribed the API alone. Anyway, it also returns the previous fields with the same name, so whoever uses will continue to work for them.

The rest of the changes if I am not wrong are in the API documentation (still working on it).

> As for the proposal, I'm not concerned much about breaking the original API by requiring an extra header. Alternatively, for some time, we can assume that no header with version selector means v1. The reason why I'm not very much a fan of `v2` in the path is that I would like to start with a clean slate and not to keep in the path the heritage of originally having two APIs.
> 
> We can easily have two different API routers (for v1 and v2) and just decide between them based on the header parameter which will specify the version (and default to v1 for the time being).
> 

Ok, i already edited the draft https://github.com/Mailtrain-org/mailtrain/pull/935#issuecomment-678628497

> I like the idea with the V2 API page. Just as I said, I would like to go with the paths /api/xxxx and specify the version in the header. It will also allow us to accommodate breaking changes in the API later on.

If you could upload a skeleton with the routes, which processes the request header, a blank page for APIv2.js and an example of using curl with request header, I'll fill it in.

bures:
Perfect. Thanks. Let's move the new API to a separate issue. I'll provide the routes and the request header with the curl example later today or tomorrow.

PR #937, opened 2020-08-24
Dates improvements (v2)

joker-x:


bures:
Great. I assume you tested it, right? Thanks.

joker-x:
Yes, with mariadb 10.5.5

Works edit, import and yyyy-mm-dd format:
imagen

Malm:
Looking forward to this fix, is there any reason why it has not been merged yet?

server/lib/knex.js
```diff
@@ -17,6 +17,15 @@ const knex = require('knex')({
             'DATETIME'
         ]
     },
+    pool: {
+        min: 2,
+        max: 10,
+        afterCreate: function(conn, cb) {
+            conn.query('SET sql_mode="ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION";', function (err) {
```
bures, line 24: What is the reason for this? Could we do it without? I don't wont to include stuff which does not address any particular problem. Thanks.

joker-x: Since MySQL 5.7 STRICT_TRANS_TABLES is enabled by default, you can disable globally in my.cnf or in database connection. For docker in this issue https://github.com/Mailtrain-org/mailtrain/issues/748 fix removing all sql_modes. I prefer remove only STRICT_TRANS_TABLES and keep the others default values ERROR_FOR_DIVISION_BY_ZERO, NO_AUTO_CREATE_USER, NO_ENGINE_SUBSTITUTION. I thing that adding in afterCreate function of knex connection is the better place to solve it.

joker-x: With STRICT_TRANS_TABLES enabled, mailtrain do not save any field of type date or timestamp, throwing ER_TRUNCATED_WRONG_VALUE

bures: But why is it? As far as I understand, It disables checking, but we should not fill in invalid data anyway.

joker-x: I think that if all inserts and updates will be called with 'YYYY-MM-DD HH:mm:ss' format in UTC should works without database errors, but i did not find where are in source code. This workaround is better that reconfigure manually the database for date fields begin to works and it can be easily remove if we find another solution.

bures: I think we have to find the problem. In my view, disabling the check just masks the problem.

joker-x: image
Maybe change .toISOString() in subscriptions models to .format("YYYY-MM-DD HH:mm:ss") can be solve it...

Try?

joker-x: Seems to works! :) In a while push the fix to this PR

joker-x:
Can anybody test if works??

joker-x:
Before you change from .toDate() to toISOString() in this commit https://github.com/Mailtrain-org/mailtrain/commit/b37ad9863c9d5b0ea38f440f684102467fa8dbe3#diff-9a370bff7587e5d2f36495f32f4101be

Can it break triggered campaigns?

bures:
Yep. This looks good. @joker-x - let me know once it is ready to be merged.

@Malm - can you test it?

Thanks.

bures:
@joker-x Honestly, I'm not sure. It has been quite some time since I made the code. 

As far as triggered campaigns are concerned, I'm not even sure if they work properly now. I don't use them and I wonder if anyone does.

joker-x:
Ok, @bures ... step by step, if the same error is present in other place, now we known how to solve it, and 'moment' is a dependency in server and client side

In my test site is working, but i tuned several configurations in the server, not a clean installation, better if anyone else can test it too.

joker-x:
Why save all date types as timestamp in database?

joker-x:
I tested in a clean installation and it works! Can you merge it @bures ?

bures:
Perfect. Thanks.

Issue #810, opened 2019-12-17
VERP failing when campaign cid starts with "-"


priethor:
When one of my campaigns cid starts with a hyphen, if I activate VERP in the send configuration, some recipient servers reject the mail because of un illegrall address syntax of the sender:

> Dec 16 09:12:31 server postfix/smtpd[32636]: warning: Illegal address syntax from XXX.com[A.B.C.D] in MAIL command: <-NR_vvx2w.r1NMEQeyX.HJ-Q4mlJQ@XXX.com>

We have checked many possibilities and it just happens with VERP and when the campaign cid starts with "-", as just by cloning the affected campaigns solves the issue.

There are 2 ways I can see to fix this issue:
- Not allowing newly created campaign cids to start with "-", regenerating them if requried.
- Always adding to the VERP mail addresses some static prefix, but this might not be as easy as it sounds if it is fully handled by ZoneMTA.

pangea-org:
Confirmed in here too, successfully reproduced.
Mailtrain log says : 
  
Error: Mail command failed: 501 5.1.7 Bad sender address syntax

priethor:
This bug is currently reproducing in other parts of the application, such us links failing when the campaign starts with "-", although I cannot reproduce in a consisten way (links don't always fail)

forgeOneHQ:
We are going to start with the development and testing of mailtrain v3 in the next weeks.

You are welcome to help us with the testing as soon as the first release candidate is available.

PR #938, opened 2020-08-25
Reemplace shortid with nanoid (v2)

joker-x:
- shortid is deprecated, its developers recommend nanoid that is faster and safe
- to minimize changes in code, i write a module in lib/shortid.js that implement the functions generate and isValid with nanoid
- the alphabet and length of cid can be set in configuration file
- the default value of alphabet is 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ', i remove numbers and dash to avoid errors as https://github.com/Mailtrain-org/mailtrain/issues/810
- the default value of length of cid is 10


bures:
Perfect. Looks good. Thanks.

simevidas:
FWIW, I’m trying to set up Mailtrain on FastComet’s shared hosting. There’s an error caused by shortid:

> You required the Node Cluster module, which is incompatible with Passenger…

I will try to add the above code change to my Mailtrain v1 project and see if that fixes it. Wish me luck.

Issue #940, opened 2020-08-27
(v2) Question: Hide Menu Items


elitzer2:
Hello! I see that the Reports menu item can be disabled but it seems like the others are hardcoded.

1. I would like to remove the Channels menu item as we won't be using Channels.
 a) Are Channels really for pre-filling some information that you'd otherwise manually fill out in a regular campaign? Meaning if I go to create a regular campaign and choose a Channel should it pre-fill the information that's in the Channel? Not sure if there's a bug or if I'm not entirely understanding the purpose/use of a Channel.
2. Additionally, I'd like to remove the API under administration as I don't want normal users to be able to use/generate API tokens.
3. I would also like to remove the Triggered/RSS options for campaigns for the time being.

Any thoughts on how I could do these? Thanks!

joker-x:
Have you try to remove 'view' permission over 'channel' entity for the rol of the users?

elitzer2:
> Have you try to remove 'view' permission over 'channel' entity for the rol of these users?

Yeah I did try removing the createChannels from [here](https://github.com/Mailtrain-org/mailtrain/blob/development/server/config/default.yaml#L316) and [here](https://github.com/Mailtrain-org/mailtrain/blob/development/server/config/default.yaml#L327) as well as making the channel config itself an empty array [here](https://github.com/Mailtrain-org/mailtrain/blob/development/server/config/default.yaml#L327). Looking at the React code there doesn't seem to be a way to define the channel menu item via config and doesn't seem like it's dynamic based on a channel view permission.

forgeOneHQ:
We are going to start with the development and testing of mailtrain v3 in the next weeks.

You are welcome to help us with the testing as soon as the first release candidate is available.

PR #942, opened 2020-08-27
Added new global permissions to manage main entities, hide menu entry and deny access if not have the permission (v2)

joker-x:
Added new global permissions: **manageLists, manageChannels, manageTemplates, manageCampaigns, manageReports, manageApi, manageSendConfigurations, manageNamespaces** to already defined **manageBlacklist, manageSettings** and change the name of **displayManageUsers** to **manageUsers**

If not assign manageChannels to the rol of user, the channel menu will be hide, for example.

Also, if user is a superadmin (he has the admin:true in his configuration) always show all menu entries.

### DIFF
--- a/client/src/root.js
+++ b/client/src/root.js
@@ -73,7 +73,7 @@ class Root extends Component {
 		                const entry = topLevelItems[entryKey.toLowerCase()];
 		                const link = entry.link || entry.externalLink;
 
-		                if (mailtrainConfig.user.admin || mailtrainConfig.globalPermissions["manage"+entryKey]) {
+		                if (mailtrainConfig.user.admin || mailtrainConfig.globalPermissions["displayManage"+entryKey]) {
 		                    if (link && path.startsWith(link)) {
 		                        topLevelMenu.push(<NavLink key={entryKey.toLowerCase()} className="active" to={link}>{entry.title} <span className="sr-only">{t('current')}</span></NavLink>);
 		                    } else {
@@ -87,17 +87,17 @@ class Root extends Component {
                             <ul className="navbar-nav mt-navbar-nav-left">
                                 {topLevelMenu}
                                 <NavDropdown label={t('administration')}>
-                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.manageUsers) &&
+                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.displayManageUsers) &&
                                         <DropdownLink to="/users">{t('users')}</DropdownLink>}
-                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.manageNamespaces) &&
+                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.displayManageNamespaces) &&
                                         <DropdownLink to="/namespaces">{t('namespaces')}</DropdownLink>}
                                     {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.manageSettings) &&
                                         <DropdownLink to="/settings">{t('globalSettings')}</DropdownLink>}
-                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.manageSendConfigurations) &&
+                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.displayManageSendConfigurations) &&
                                         <DropdownLink to="/send-configurations">{t('sendConfigurations')}</DropdownLink>}
                                     {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.manageBlacklist) &&
                                         <DropdownLink to="/blacklist">{t('blacklist')}</DropdownLink>}
-                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.manageApi) &&
+                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.displayManageApi) &&
                                         <DropdownLink to="/account/api">{t('api')}</DropdownLink>}
                                 </NavDropdown>
                             </ul>
--- a/client/src/root.js
+++ b/client/src/root.js
@@ -70,13 +70,12 @@ class Root extends Component {
                 if (mailtrainConfig.isAuthenticated) {
 
                     const gP = mailtrainConfig.globalPermissions;
-                    const superadmin = mailtrainConfig.user.admin;
 
 		            for (const entryKey of topLevelMenuKeys) {
 		                const entry = topLevelItems[entryKey.toLowerCase()];
 		                const link = entry.link || entry.externalLink;
 
-		                if (superadmin || gP["manage"+entryKey]) {
+		                if (gP["manage"+entryKey]) {
 		                    if (link && path.startsWith(link)) {
 		                        topLevelMenu.push(<NavLink key={entryKey.toLowerCase()} className="active" to={link}>{entry.title} <span className="sr-only">{t('current')}</span></NavLink>);
 		                    } else {
@@ -89,20 +88,20 @@ class Root extends Component {
                         <>
                             <ul className="navbar-nav mt-navbar-nav-left">
                                 {topLevelMenu}
-                                {(superadmin || gP.manageUsers || gP.manageNamespaces || gP.manageSettings || 
+                                {(gP.manageUsers || gP.manageNamespaces || gP.manageSettings || 
                                   gP.manageSendConfigurations || gP.manageBlacklist || gP.manageApi) &&
                                   <NavDropdown label={t('administration')}>
-                                    {(superadmin || gP.manageUsers) &&
+                                    {(gP.manageUsers) &&
                                         <DropdownLink to="/users">{t('users')}</DropdownLink>}
-                                    {(superadmin || gP.manageNamespaces) &&
+                                    {(gP.manageNamespaces) &&
                                         <DropdownLink to="/namespaces">{t('namespaces')}</DropdownLink>}
-                                    {(superadmin || gP.manageSettings) &&
+                                    {(gP.manageSettings) &&
                                         <DropdownLink to="/settings">{t('globalSettings')}</DropdownLink>}
-                                    {(superadmin || gP.manageSendConfigurations) &&
+                                    {(gP.manageSendConfigurations) &&
                                         <DropdownLink to="/send-configurations">{t('sendConfigurations')}</DropdownLink>}
-                                    {(superadmin || gP.manageBlacklist) &&
+                                    {(gP.manageBlacklist) &&
                                         <DropdownLink to="/blacklist">{t('blacklist')}</DropdownLink>}
-                                    {(superadmin || gP.manageApi) &&
+                                    {(gP.manageApi) &&
                                         <DropdownLink to="/account/api">{t('api')}</DropdownLink>}
                                   </NavDropdown>}
                             </ul>
--- a/client/src/root.js
+++ b/client/src/root.js
@@ -76,7 +76,7 @@ class Root extends Component {
 		                const entry = topLevelItems[entryKey.toLowerCase()];
 		                const link = entry.link || entry.externalLink;
 
-		                if (superadmin || gP["displayManage"+entryKey]) {
+		                if (superadmin || gP["manage"+entryKey]) {
 		                    if (link && path.startsWith(link)) {
 		                        topLevelMenu.push(<NavLink key={entryKey.toLowerCase()} className="active" to={link}>{entry.title} <span className="sr-only">{t('current')}</span></NavLink>);
 		                    } else {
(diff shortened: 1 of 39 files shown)


joker-x:
Fix: #940

bures:
Looks good. But can we keep the distinction between `manageXXX` and `displayManageXXX`? 

The problem is that manageBlacklist and manageSettings are really permissions which are checked in `models/blacklist.js` and `models/settings.js` respectively. On the other hand `displayManageUsers` determines only the visibility of the menu item. It's not checked in the models. This means that if one makes a REST call directly, he/she can still manage the users provided one has the corresponding permission in a namespace. 

Thanks.

joker-x:
Yes, of course, rename only the new ones? rename manageSettings and manageBlacklist too?

elitzer2:
This is great!

elitzer2:
This probably isn't the best place to ask but since it's sort of related and I'm curious, are there steps for contributors to build the node locally? I haven't worked with a handlebars app before. Thanks!

client/src/root.js
```diff
@@ -67,29 +67,38 @@ class Root extends Component {
...
+                                        <DropdownLink to="/namespaces">{t('namespaces')}</DropdownLink>}
+                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.manageSettings) &&
+                                        <DropdownLink to="/settings">{t('globalSettings')}</DropdownLink>}
+                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.displayManageSendConfigurations) &&
+                                        <DropdownLink to="/send-configurations">{t('sendConfigurations')}</DropdownLink>}
+                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.manageBlacklist) &&
+                                        <DropdownLink to="/blacklist">{t('blacklist')}</DropdownLink>}
+                                    {(mailtrainConfig.user.admin || mailtrainConfig.globalPermissions.displayManageApi) &&
+                                        <DropdownLink to="/account/api">{t('api')}</DropdownLink>}
                                 </NavDropdown>
```
elitzer2, line 102: Would you also mind adding a check that if none of the items under Administration are present, the top-level Administration label is also hidden?

joker-x: Ready ;)

elitzer2: Lovely!

joker-x:
> This probably isn't the best place to ask but since it's sort of related and I'm curious, are there steps for contributors to build the node locally? I haven't worked with a handlebars app before. Thanks! I suppose you might just use the develop Docker container and it points to the node! I was likely over-thinking...
> 
> UPDATE: [I can't read...](https://github.com/Mailtrain-org/mailtrain/tree/development#deployment-with-docker-and-docker-compose-for-development). I am not running our production instance using Docker so I was trying to figure out how I can make local changes and build the node. But I should be able to figure that out from the steps provided and if not, look into the docker container commands.

I don't use docker, to apply these changes when approved you can do something similar steps:
```
cd <REPOSITORY_PATH>
git pull origin development
# edit server/config/production.yml to write the new global permissions
cd client
NODE_ENV=production npm run build
service mailtrain restart #as root
```

elitzer2:
> > This probably isn't the best place to ask but since it's sort of related and I'm curious, are there steps for contributors to build the node locally? I haven't worked with a handlebars app before. Thanks! I suppose you might just use the develop Docker container and it points to the node! I was likely over-thinking...
> > UPDATE: [I can't read...](https://github.com/Mailtrain-org/mailtrain/tree/development#deployment-with-docker-and-docker-compose-for-development). I am not running our production instance using Docker so I was trying to figure out how I can make local changes and build the node. But I should be able to figure that out from the steps provided and if not, look into the docker container commands.
> 
> I don't use docker, to apply these changes when approved you can do something similar steps:
> 
> ```
> cd <REPOSITORY_PATH>
> git pull origin development
> # edit server/config/production.yml to write the new global permissions
> cd client
> NODE_ENV=production npm run build
> service mailtrain restart #as root
> ```

Awesome, thanks! I actually added your repo as a remote and have been using your branch so I can start configuring our new instance. We are moving away from Mailchimp and I *really* want Mailtrain to serve our basic needs as a small nonprofit, and I'm happy because I think it will.

joker-x:
@bures now the menu is hidden if the user do not have the global permission, but the user still can access by the url of the resource, i think would be better enforce the global permission as blacklist do and if true later check particular permission.

bures:
@joker-x - I agree. It would indeed be better to enforce the permissions. I didn't suggest it in the first place because it needs quite some work because the enforcement has to go to many places in models. But if you like to do it, I believe it would be the best option.

To recap:
- drop existing `displayManageXXX` permissions in favor `manageXXX`
- introduce new `manageXXX` permissions
- check the `manageXXX` permissions in models, such that it is really about access, not just about showing it in client

Thanks a lot.

server/models/shares.js
```diff
@@ -449,7 +451,7 @@ async function regenerateRoleNamesTable() {
 
 
 function throwPermissionDenied() {
-    throw new interoperableErrors.PermissionDeniedError('Permission denied');
+    throw new interoperableErrors.PermissionDeniedError(tMark('permissionDenied'));
```
joker-x, line 454: don't work :( i also tried with tUI but i don't know how to define locales param

joker-x:
Seems to work, need more testing, can we help us @elitzer2 ?

server/models/shares.js
```diff
@@ -451,7 +451,7 @@ async function regenerateRoleNamesTable() {
 
 
 function throwPermissionDenied() {
-    throw new interoperableErrors.PermissionDeniedError(tMark('permissionDenied'));
+    throw new interoperableErrors.PermissionDeniedError(tUI('permissionDenied', config.defaultLanguage));
```
joker-x, line 454: I don't know how to pass the current user language selected in locale param

joker-x:
Even though the checkGlobalPermission function returns true if context.user.admin, it denies access if you don't have global permission:

https://github.com/Mailtrain-org/mailtrain/blob/development/server/models/shares.js#L497

elitzer2:
> Seems to work, need more testing, can we help us @elitzer2 ?

Sure I can take a peek over the next couple of days and poke around.

elitzer2:
One thing right off the bat I noticed is that if I disable `manageChannels` I can't create campaigns due to permissions errors. 

Here's my custom config:
```
defaultRoles:
  global:
    campaignsAdmin:
      name: Campaigns Admin
      description: Under the namespace in which the user is located, the user has all permissions for managing lists, templates and campaigns and the permission to send to send configurations.
      permissions: [setupAutomation, manageLists, manageTemplates, manageCampaigns]
      ownNamespaceRole: campaignsAdmin

  namespace:
    campaignsAdmin:
      name: Campaigns Admin
      description: In the respective namespace, the user has all permissions for managing lists, channels, templates and campaigns and the permission to send to send configurations.
      permissions: [view, edit, delete, createList, createTemplate, createCampaign]
      children:
        sendConfiguration: [viewPublic, sendWithoutOverrides, sendWithAllowedOverrides]
        list: [view, edit, delete, viewFields, manageFields, viewSubscriptions, viewTestSubscriptions, manageSubscriptions, viewSegments, manageSegments, viewImports, manageImports, send, sendToTestUsers]
        customForm: []
        campaign: [view, edit, delete, viewAttachments, manageAttachments, send, sendToTestUsers, viewStats]
        template: [view, edit, delete, sendToTestUsers]
        channel: []
        mosaicoTemplate: [view, viewFiles]
        namespace: [view, edit, delete, createList, createTemplate, createCampaign]
```

elitzer2:
Idk the permissions are a bit confusing to me and seem to collide a bit sometimes. For example, I wanted to get rid of `viewFiles`/`manageFiles` from campaign & template but then I would get permissions errors when cloning campaigns because it would try and view the files of the campaign I am guessing even though there were none. I just don't want people to have the ability to see the Files option or upload Files. I might just need to keep playing around with different combinations.

server/models/shares.js
bures: I'm afraid, it is not supposed to work. The exceptions are not supposed to be localized since they are not the way we communicate with end-users. Exception is and should be the point when we or the administrator failed (e.g. in setting the permissions). 

The tMark is used to mark the point where the translation is used in the code so that the automated script (extract.js) can identify it and transform it into string resource id (e.g. 'permissionDenied').
The tUI is used for generated translated strings on the server. In this case, however, the language has to be given as a parameter because the server does not assume language implicitly.

bures:
I think this is getting out of hand. I fear we started patching the permission system without having clear use-cases. The features of Mailtrain are intertwined. You cannot disable one without impacting the others. Similarly, the permissions are not orthogonal. There are only certain combinations that work.

Looking at all this, I think we should not mix the permissions and the fact that some option is enabled in the UI. It's just too complex to handle it and typically results in some error. 

> > Seems to work, need more testing, can we help us @elitzer2 ?
> 
> Sure I can take a peek over the next couple of days and poke around.

Not sure what you mean. What exactly does not work?

elitzer2:
> Looking at all this, I think we should not mix the permissions and the fact that some option is enabled in the UI. It's just too complex to handle it and typically results in some error.

Yeah, agreed. Separation of concerns is good. If getting rid of UI options adds too much complexity to the config I can just take care of it on my end for my use case.

bures (commented):
I'm afraid this is not something I'm happy about. I would like to avoid any explicit super admin as long as possible. The 'admin' attribute in the permission config just says which role should be given to user no. 1. It serves as a bootstrap for permissions to fix cases when by whatever error the user no. 1 looses his/her permissions.

bures (commented):
I'm afraid I'm not happy with using the 'admin' or 'superadmin' to determine what is supposed to be shown in the menu. As I wrote in the other comment.

bures:
@elitzer2 - Please suggest which options you want to hide from the menu, for which users and why (a bit of rationale so that we understand your case in the context). We can then figure out some better way to customize the UI based on user role (either via 'displayManageXXX' permissions or via another way).

elitzer2:
> @elitzer2 - Please suggest which options you want to hide from the menu, for which users and why (a bit of rationale so that we understand your case in the context). We can then figure out some better way to customize the UI based on user role (either via 'displayManageXXX' permissions or via another way).

I will answer the inverse since there are fewer options, but I want a user that only has Lists, Templates, and Campaigns as menu items. I will be configuring the rest on behalf of the users and I want them to have a simplified interface with the least amount of control, mostly just to reduce confusion with all of the options (which is a good thing, but sometimes can be overwhelming to some).

bures:
@elitzer2 - Does it mean you want to have all the settings under lists (custom fields, segments, triggers, etc.) and similarly under templates and campaigns?

Another questions, out of curiosity, why do you want templates? I adopted them because they were in v1, but when using Mailtrain, I could not find much use for templates. I have users use only campaigns. Out of templates, I use only Mosaico templates.

elitzer2:
> @elitzer2 - Does it mean you want to have all the settings under lists (custom fields, segments, triggers, etc.) and similarly under templates and campaigns?
> 
> Another questions, out of curiosity, why do you want templates? I adopted them because they were in v1, but when using Mailtrain, I could not find much use for templates. I have users use only campaigns. Out of templates, I use only Mosaico templates.

I think I can live with all the options that are on those pages. Would be a nice to have to be able to disable some of those but not a requirement of mine. Only things I can think of I'd want to get rid of is the RSS & Triggered Campaigns and the Files UI button so users can't upload any Files there. I guess getting rid of the Mosaico Templates button on the Templates page would be good too. 

The reason I can see people using templates is having different formats depending on the email type (quick send of a small blurb, event flyer, etc.) but I am sure eventually they will end up cloning campaigns since it prefills all the lists/send configuration for you (Is that what a Channel is supposed to do? I still don't understand those yet). I was thinking of having them use the "Custom content cloned from template" option and using the templates as a base with blocks/layout vs. fully-fledged emails. And I agree, the only editors I enabled is Mosaico. It should be sufficient for our needs.

joker-x:
Ok, I still thinking that the menu entries and buttons must be show only if you have the relative permission and the role marked with admin = true must have all permissions allthough don't set in the config. But I stop working in this PR until we are agree with the solution.

Happy weekend!

bures:
@joker-x Great. Let's bring this back to the drawing table and see what the right solution is. Thanks for your effort. Happy weekend to you too.

bures:
@elitzer2 - The channels are to group series of emails on the same topic. An example is a though of a day or weekly newsletter. The channel allows you to show only campaigns sent within the channel. It also contains defaults that make it easy to create another campaign in the channel. 

Cloning a campaign is another way, but does not give you the benefit of seeing only campaigns logically belonging together (like all the thoughts of the day) and it is more error prone to copy-paste errors. I rather persuade my users to always start a campaign with clean slate (i.e. with pre-filled fields by the channel and customized mosaico template).

elitzer2:
@bures Ah ok that might be what I'm looking for and instead of creating a campaign on the Campaigns page I can have them create it through the Channel so things are pre-filled that way. I'm the one training them on how to use it, so it's still up in the air in terms of the final flow. I'll dig into it a bit more, thanks!

bures:
> @bures Ah ok that might be what I'm looking for and instead of creating a campaign on the Campaigns page I can have them create it through the Channel so things are pre-filled that way. I'm the one training them on how to use it, so it's still up in the air in terms of the final flow. I'll dig into it a bit more, thanks!

Yep. That was the idea behind channels. Actually after channels were introduced, the "Campaigns" could be made a bit less prominent in the menu because they are kind of like "All campaigns including RSS and triggered ones and those without a channel assigned".

joker-x:
> One thing right off the bat I noticed is that if I disable `manageChannels` I can't create campaigns due to permissions errors.
> 
Works with the last commit, i think

server/models/shares.js
joker-x: We use https://github.com/smhg/express-locale that add to req the language, perhaps we can extract it from request.

bures: You Are right that we could possibly extract it from the cookie sent together with the REST request. However, it does not really change much on the fact that "Permission denied" is not something the end-user should see. The UI should not offer choices the user has not permissions for. If the permission denied can be obtained, then the client should catch the exception and show it in a more user-friendly way. As such, I would really not localize the exceptions. Exceptions are for logs, not for end-users.

joker-x: Can we do the translation in flash notice only?
imagen

bures: Sure. Only it should not be translated by the server, but by the client.

Check client/src/lib/modals.js:164. I belive you could add something similar to general error handling in client/src/lib/page.js:461. There I would check for the `error instanceof interoperableErrors.PermissionDeniedError` and would present a fitting error message possibly with giving some hints (like someone may have changed the permissions while you worked, if you think it should work, contact the administrator or similar).

bures:
@joker-x - Thanks for the effort. But to be honest upfront, I'm not sure I will accept this PR. It's getting too complicated and I'm not sure about the necessity the "permission denied" check in the component mount. Given the problems with "manageChannels", I'm also not convinced we should use manageXXX permissions to control what one sees in the UI. Due to how the things are intertwined, some of the methods (typically getters) cannot be protected by the global manageXXX permission, which casts a doubt on what the semantics of manageXXX is and if they are useful at all since they don't fully protect the API. At this point, I'm veering towards leaving the permission system as-is and having a dedicated section in the config, which determines what a user should see depending on its role. In my view, it would be more flexible, because it would also allow things like rearranging the menu order, discarding "Campaigns" top-level menu a bringing it under Channels as a "All campaigns" button, etc. 

Sorry, if I misled you anyhow previously. I'm also making my mind as I see the progress and the issues that arise on the way.

Issue #944, opened 2020-08-28
(v2) Embedded subscription widget working?


elitzer2:
It seems like the embedded widget has been ported into the v2 codebase, is it in working condition? It's the last piece I need to be able to fully convert over to Mailtrain. I tried the following code and it doesn't seem to be working (found it on an external site). I have added the cors configuration already.

```
<div data-mailtrain-subscription-widget data-url="https://lists.domain.com/subscription/xciBjhXRN/widget">
     <a href="https://lists.domain.com/subscription/xciBjhXRN">Subscribe to our list</a>
</div>
<script src="https://mailtrain.domain.com/subscription/widget.js"></script>
```

elitzer2:
Ok, so totally missed the instructions [here](https://github.com/Mailtrain-org/mailtrain/tree/development/docs#subscription-widget)... And onto my next problem!

<img width="1503" alt="Screen Shot 2020-08-28 at 10 57 44 PM" src="https://user-images.githubusercontent.com/5624305/91628070-e3cae780-e981-11ea-8eee-5c5becabda96.png">

elitzer2:
OK totally missed the instructions [here](https://github.com/Mailtrain-org/mailtrain/tree/development/docs#subscription-widget) and onto my next problem... Email field label not rendering and getting browser CORS errors when submitting.

<img width="1514" alt="Screen Shot 2020-08-28 at 10 59 20 PM" src="https://user-images.githubusercontent.com/5624305/91628104-42906100-e982-11ea-8cdb-572d1fbdd7a1.png">

elitzer2:
Alright I got it working but I'm not sure it's the "right way". Need some help with setting up CORS correctly before I try and PR it. I was able to get it working by adding `router.use(cors(corsOptions));` [here](https://github.com/Mailtrain-org/mailtrain/blob/development/server/routes/subscription.js#L49) and changing [this](https://github.com/Mailtrain-org/mailtrain/blob/development/server/routes/subscription.js#L328) to `subscribeUrl: getPublicUrl('subscription/${list.cid}/subscribe')`.

I'm afraid that this turns on CORS for all routes but seems like we just want them for two routes. Any advice?

elitzer2:
Ok I've got it working just need to change `getTrustedUrl` to `getPublicUrl` on [this line](https://github.com/Mailtrain-org/mailtrain/blob/development/server/routes/subscription.js#L328).

PR #948, opened 2020-08-29
(v2) Fix subscription widget

elitzer2:
Fixes the subscription widget! Had an invalid URL and CORS doesn't allow redirects so needed to specify a JSON API response in the case of an existing subscription on an email.

I wasn't sure if I was supposed to use `tMark` or `tUI` here, what's the main difference in use? What happens when a translation doesn't exist for other languages? Do we have to add translations to each language?

Closes https://github.com/Mailtrain-org/mailtrain/issues/944

This also fixes the Email field label not appearing! Needed to use the template rendering to handle the translations.

### DIFF
--- a/server/routes/subscription.js
+++ b/server/routes/subscription.js
@@ -331,14 +331,15 @@ router.getAsync('/:cid/widget', cors(corsOptions), async (req, res) => {
         subscribeUrl: getPublicUrl(`subscription/${list.cid}/subscribe`),
         hasPubkey: !!configItems.pgpPrivateKey,
         customFields: await fields.forHbs(contextHelpers.getAdminContext(), list.id),
-        template: {},
+        template: 'subscription/widget-subscribe.hbs',
         layout: null,
     };
 
     await injectCustomFormData(req.query.fid || list.default_form, 'web_subscribe', data);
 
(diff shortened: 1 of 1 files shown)


server/routes/subscription.js
```diff
@@ -281,6 +281,9 @@ router.postAsync('/:cid/subscribe', passport.parseForm, corsOrCsrfProtection, as
 
     if (existingSubscription && existingSubscription.status === SubscriptionStatus.SUBSCRIBED) {
         await mailHelpers.sendAlreadySubscribed(req.locale, list, email, existingSubscription);
+        if (req.xhr) {
```
bures, line 284: I'm afraid this should not be here. If I understand correctly, It essentially allows anyone to check if an email is in the mailing list.

elitzer2: Hmm. Yeah you're right on that, good call. We have to return some sort of JSON though or it breaks the widget. Can we just return a success then? That way through the API you won't really know.
```
return res.status(200).json({
    msg: tUI('pleaseConfirmSubscription', req.locale)
});
```

bures:
Yep. Success would be just right. The flow should be the same as in regular subscriptions. It should just say that email was sent and that the user should follow the instructions there.

bures:
Good. Thanks.

Issue #927, opened 2020-07-30
(V2) Channels does not correctly pre-fill "Template" section


Britaliope:
### Description
When creating a channel, we have the possibility to fill the "Template" section, but: 

- When choosing "Custom content cloned from template" and choose a template, the choosen template is not selected when creating a campaing from the channel page. If we then try to edit the channel, the field used to choose a template is not shown.
- When choosing "Template" and choose a template, the choosen template is not selected when creating a campaing from the channel page. But this time, the choosen template is correctly shown in the Channel edit page. 

The option "Custom content cloned from another campain" works as expected
 
### Mailtrain version
`develop` branch, last commit: `commit 8dab13d903e8f19e8d69887977993e84c7768b0b` (Sat Jul 25 10:24:34 2020)

### Steps to reproduce
- Create a template and a channel with "Custom content cloned from template" as Content source and choose the previously created template. 
- Create a campaign from this channel page (mailtrain-url.tld/channels/<id>/campaigns)
- at the bottom of the page, **The choosen template is not pre-filled**
- Go to edit channel page (mailtrain-url.tld/channels/<id>/edit)
- The "template" field have disappeared

elitzer2:
@bures I went to go try and set up Channels like we discussed and ran into this issue. Haven't dug into it at all yet.

elitzer2:
Fixed. Can close this issue.

bures:
Perfect. Thanks.

PR #949, opened 2020-08-31
Fixes the autofill of campaign from cloned template when created via channel

elitzer2:
Fixes https://github.com/Mailtrain-org/mailtrain/issues/927


bures:
Looks good. Thanks.

Issue #850, opened 2020-02-13
html-to-text is not working


Akatana:
Hi, 
since the newest update I am no longer able to convert HTML to text.
I am getting a 500 Error code because /editorapi/html-to-text?editor=mosaico is not accessible.
I also tried the other editors but none is working.

Thanks for the help
Akatana

gustavdz:
I have the same issue when I try to get the text from the HTML I got this error: "Unexpected token < in JSON at position 0"

elitzer2:
Looks like they took down the API, see https://github.com/Mailtrain-org/mailtrain/issues/943

bures:
Yeah. I guess they did. Any fix would be welcome.

elitzer2:
Fixed. Can close this issue.

Issue #943, opened 2020-08-28
(v2) Plain text rendering of Mosaico template issue


elitzer2:
Not sure if I've set something up incorrectly, but when I try to view the plain text of the email using the "To extract the text from HTML click here" link, I get the following stack dump. I am running on `Red Hat Enterprise Linux release 8.2` with `premailer-api@1.0.4` on an EC2 instance and ran the CentOS 8 install script to get things going following [these](https://github.com/Mailtrain-org/mailtrain/tree/development#installation-on-fresh-centos-7-or-ubuntu-1804-lts-public-website-secured-by-ssl) instructions. I noticed the `mvis` client/server folders don't have a `package-lock.json` defined so not sure if it would be due to any differences in versions in there.

There are the template settings used in the campaign:
<img width="491" alt="Screen Shot 2020-08-27 at 7 38 04 PM" src="https://user-images.githubusercontent.com/5624305/91508361-f3243500-e89c-11ea-9df7-6f939f3d4c79.png">

I also just tested this in a local docker container and it seems to be broken there as well.

```
node[4763]: verb HTTP SyntaxError: Unexpected token < in JSON at position 0
node[4763]: verb HTTP     at JSON.parse (<anonymous>)
node[4763]: verb HTTP     at Request.handlePremailerResponse [as _callback] (/opt/mailtrain/server/node_modules/premailer-api/premailer.js:103:23)
node[4763]: verb HTTP     at Request.self.callback (/opt/mailtrain/server/node_modules/request/request.js:185:22)
node[4763]: verb HTTP     at Request.emit (events.js:198:13)
node[4763]: verb HTTP     at Request.EventEmitter.emit (domain.js:448:20)
node[4763]: verb HTTP     at Request.<anonymous> (/opt/mailtrain/server/node_modules/request/request.js:1161:10)
node[4763]: verb HTTP     at Request.emit (events.js:198:13)
node[4763]: verb HTTP     at Request.EventEmitter.emit (domain.js:448:20)
node[4763]: verb HTTP     at IncomingMessage.<anonymous> (/opt/mailtrain/server/node_modules/request/request.js:1083:12)
node[4763]: verb HTTP     at Object.onceWrapper (events.js:286:20)
node[4763]: verb HTTP     at IncomingMessage.emit (events.js:203:15)
node[4763]: verb HTTP     at IncomingMessage.EventEmitter.emit (domain.js:448:20)
[...]

PR #950, opened 2020-08-31
Fix plaintext rendering

elitzer2:
Uses the plaintext rendering method used for sending out campaigns that [already existed](https://github.com/Mailtrain-org/mailtrain/blob/development/server/lib/message-sender.js#L722) in the app since Premailer API was shut down.

Fixes https://github.com/Mailtrain-org/mailtrain/issues/850 and https://github.com/Mailtrain-org/mailtrain/issues/943


bures:
Thanks.

PR #953, opened 2020-09-01
(v2) Required custom fields (client side validation only)

elitzer2:
Alright, so let me know what you think about putting this in for now with just client side validation. I know it's not ideal but this will cover my use case for the most part as none of my users will be altering page content. Eventually if I have time I'd like to add in server validation, or someone else can as well as the data needed will be present in the DB it just needs to be checked someplace.

HTML5 handles it all for free so this was a quick way to handle the client side. Not sure if there is a library or mechanism on the server to check form validity generally but I imagine if I tried to write it myself there would be a lot of different checks based on the input type. I included the "client side validation only" in the language displayed to the User so it's immediately obvious.

For the server, not sure if one of these might work?
https://github.com/flowstudio/datalize
https://github.com/express-validator/express-validator

If you feel it's best to wait until we have client + server side validation, I would understand. I'm just not sure of the complexity of it yet and how much time I'll have.

Thoughts? Thanks!

FYI: I tested all the input types, including the grouped option ones. It gets a little funky with the grouped ones because sometimes you want to set it on the parent group field but for radio buttons for example it has to be set on at least one of the children based on how the HTML is set up. From the HTML spec: `To avoid confusion as to whether a radio button group is required or not, authors are encouraged to specify the attribute on all the radio buttons in a group`, which is how I set it up.

### DIFF
--- a/client/static/subscription/form-input-style.css
+++ b/client/static/subscription/form-input-style.css
@@ -44,6 +44,16 @@ label {
     margin-left: .3em;
 }
 
+form button {
+    background: #2D3E4F;
+    color: white;
+    font-size: 16px;
+    font-weight: normal;
+    padding: 16px 24px;
+    border: none;
+    border-radius: 6px;
+    cursor: pointer;
+}
 
 /* --- Inputs ------------- */
 
--- a/server/views/subscription/web-manage.mjml.hbs
+++ b/server/views/subscription/web-manage.mjml.hbs
@@ -6,9 +6,6 @@
         <mj-text>
 {{> subscription_manage_form}}<!-- don't indent me! -->
         </mj-text>
-        <mj-button mj-class="button" href="#submit">
-            Update Profile
-        </mj-button>
         <mj-text mj-class="p">
             <a href="/subscription/{{lcid}}/unsubscribe/{{cid}}">Unsubscribe</a>
         </mj-text>
--- a/server/views/subscription/web-subscribe.mjml.hbs
+++ b/server/views/subscription/web-subscribe.mjml.hbs
@@ -6,8 +6,5 @@
         <mj-text>
 {{> subscription_subscribe_form}}<!-- don't indent me! -->
         </mj-text>
-        <mj-button mj-class="button" href="#submit">
-            Subscribe to list
-        </mj-button>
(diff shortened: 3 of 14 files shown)


server/views/subscription/web-manage.mjml.hbs
```diff
@@ -6,9 +6,6 @@
         <mj-text>
 {{> subscription_manage_form}}<!-- don't indent me! -->
         </mj-text>
-        <mj-button mj-class="button" href="#submit">
-            Update Profile
-        </mj-button>
```
elitzer2, line 11: Uses the submit button in `server/views/subscription/partials/subscription-manage-form.hbs`

server/views/subscription/web-subscribe.mjml.hbs
```diff
@@ -6,8 +6,5 @@
         <mj-text>
 {{> subscription_subscribe_form}}<!-- don't indent me! -->
         </mj-text>
-        <mj-button mj-class="button" href="#submit">
-            Subscribe to list
-        </mj-button>
```
elitzer2, line 11: Uses the subscribe to list button in `server/views/subscription/partials/subscription-subscribe-form.hbs`

client/static/subscription/form-input-style.css
```diff
@@ -44,6 +44,16 @@ label {
...
+form button {
+    background: #2D3E4F;
+    color: white;
+    font-size: 16px;
+    font-weight: normal;
+    padding: 16px 24px;
+    border: none;
+    border-radius: 6px;
+    cursor: pointer;
+}
```
elitzer2, line 56: This makes it so the button has the same styling as it had before with the `<a>` tag

elitzer2:
@bures any thoughts on this?

bures:
Thanks. I didn't test it myself, but the code looks good to me. So no problem to merge it if you tested it properly. Let me know please.

elitzer2:
I did do a bunch of testing but it could be nice if someone else could just run through it. @joker-d do you have any time to test this a bit?

elitzer2:
@bures I say we just go with it!

bures:
Thanks

PR #969, opened 2020-09-29
Fix #598

### LINKED ISSUE #598: Encoding problem with merge tags and Gmail
I'm experiencing a problem with encoding on Gmail side.

Example email content:

> Cześć [FIRST_NAME]!

Where the FIRST_NAME is "Michał". The user will see it as follows:

> Cześć Micha\&lstrok;!

The source of the email looks like follows:

```Cze=C5=9B=C4=87 Micha&lstrok;!```

The non-merge-tag characters are encoded email like, but the merge tag special character is encoded the html way and Gmail can't handle it properly. Thunderbird displays the message properly.

I've already tried changing the encoding (first line of email body) to: 

- `<meta charset="utf-8">`
- `<meta http-equiv="Content-Type" content="text/html; charset=utf-8">`
- `<meta charset="utf-8" http-equiv="Content-Type" content="text/html; charset=utf-8">`

No luck though. Any idea how to tackle this issue?

D4rk4:


galq5:
I guess this MR should have as target development branch.

Issue #855, opened 2020-02-20
ERROR: pull access denied for mailtrain, repository does not exist or may require 'docker login': denied: requested access to the resource is denied


4n70w4:
`docker-compose up`

```
Creating network "mailtrain_default" with the default driver
Creating volume "mailtrain_mailtrain-mysq-data" with default driver
Creating volume "mailtrain_mailtrain-redis-data" with default driver
Creating volume "mailtrain_mailtrain-node-data" with default driver
Creating volume "mailtrain_mailtrain-node-config" with default driver
Creating volume "mailtrain_mailtrain-node-reports" with default driver
Pulling redis (redis:3.0)...
3.0: Pulling from library/redis
f5cc0ee7a6f6: Pull complete
5fc25ed18e87: Pull complete
e025bc8872f6: Pull complete
77c68b51b836: Pull complete
7c403ece3755: Pull complete
0a653bd338f4: Pull complete
31531fd948c6: Pull complete
Digest: sha256:730b765df9fe96af414da64a2b67f3a5f70b8fd13a31e5096fee4807ed802e20
Status: Downloaded newer image for redis:3.0
Pulling mailtrain (mailtrain:latest)...
ERROR: The image for the service you're trying to recreate has been removed. If you continue, volume data could be lost. Consider backing up your data before continuing.

Continue with the new image? [yN]y
Pulling mailtrain (mailtrain:latest)...
ERROR: pull access denied for mailtrain, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

boehlke:
Hi, in the Readme I see a step saying
"Copy the file docker-compose.override.yml.tmpl to docker-compose.override.yml and modify it if you need to."
I guess you missed that point.

o-be-one:
The docker-file is 3y old... It's not working "as this".
For your error, the docker repo changed its name to mailtrain/mailtrain:latest.
Good luck!

pjsotou:
So now ask for MongoDB Server?
```
mailtrain_1  | Info: Waiting for Redis Server
mailtrain_1  | Info: Waiting for MongoDB Server
mailtrain_1  | nc: getaddrinfo: Name does not resolve
```

pjsotou:
The solution its build the image 
`docker build -t example/mailtrain:lastest .`
Then rename the image in the docker-compose.yml file to this new one

jcalfee:
change `lastest` to `latest`

[...]

PR #985, opened 2020-11-27
Clarify: docker-compose.override.yml.tmpl must be copied but only mod…

jcalfee:
…ified if needed..

https://github.com/Mailtrain-org/mailtrain/issues/855


bures:
Thanks.

PR #992, opened 2020-12-07
(V2) Force light mode in mosaico preset (iOS-Mail)

Ben-Ro:
Force iOS-Mail to use provided colours, even if the device is in darkmode to prevent unreadable mailings.

### Without fix:
image
### With fix:
image


bures:
Thanks.

PR #994, opened 2020-12-17
Update message-sender.js

ganeshkrishnan1:
catch null pointer exception. the regex returns null and crashes the whole system. also it resolves other "cannot replace" bugs raised here

### DIFF
--- a/server/lib/message-sender.js
+++ b/server/lib/message-sender.js
@@ -231,6 +231,7 @@ class MessageSender {
         }
 
         const attachments = this.attachments.slice();
+           try{
         if (replaceDataImgs) {
             // replace data: images with embedded attachments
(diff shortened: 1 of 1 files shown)


bures:
What are the steps to reproduce the error that is supposed to be fixed by this patch?

ganeshkrishnan1:
For it it was a combination of these
1) Inline image with the [ bracket
2) Tags and unsubscribe tags
3) Click on "Convert to text" below to convert the rich text to plain text automatically (weirdly some tags are double bracketed)

Send the campaigns. Either the raw base64 image or the double brackets is triggering the null pointers

bures:
That seems strange. In my view, your patch is just masking an error that might even originate somewhere else. I would rather get understanding of the error and why it arises when you click "Convert to text".

bures:

PR #997, opened 2021-01-11
Install ldap dependencies, when ldap is enabled (docker)

wagner-daniel:
With this modified docker entrypoint, the chosen ldap npm package gets installed.


MarcelWaldvogel:
Doing package installation as part of the Docker entrypoint is pretty unusual, to say the least. Can the packages not always be included as part of the docker build and only the appropriate one is used?

bures:
Looks good. Thanks.

Issue #1005, opened 2021-02-04
knex_migration fails if sql_mode is not set to ANSI_QUOTES 


manquer:
If `sql_mode` is not set with `ANSI_QUOTES`. 

The knex migration crashes at  https://github.com/Mailtrain-org/mailtrain/blob/7209e97175dee5043ad08acbcce38e43ee9cf8cc/server/lib/dbcheck.js#L72
due to the identifier being quoted with double quotes

Version: v2-beta
Branch : Development

PR #1007, opened 2021-02-04
Fixes double quote in dbCheck support when `ANSI_QUOTES`is  disabled 

manquer:
This fixes (#1005) issue with knex migration when `sql_mode` does not have `ANSI_QUOTES` included. 

[Ref MySQL 8 docs ](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes)


bures:
Thanks

PR #1052, opened 2021-04-30
Fix null template value

ericuldall:
Adding a fix for the following error:
```
    at MessageSender._getMessage (/app/server/lib/message-sender.js:254:28) 
```

The check on line 169 needed improvement


ericuldall:
Closed due to pollution from other local commits. Will resubmit

PR #1053, opened 2021-04-30
improved value check

ericuldall:
fix for
```
    at MessageSender._getMessage (/app/server/lib/message-sender.js:254:28) 
```

### DIFF
--- a/server/lib/tools.js
+++ b/server/lib/tools.js
@@ -166,7 +166,7 @@ function _formatTemplateSimple(source, mergeTags, isHTML) {
             }
         }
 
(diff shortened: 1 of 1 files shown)


bures:
@ericuldall please, can you try to change it to the following `value === undefined || value === null` and tell me if it still fixes the problem? If so, please update the PR. The current solution seems to me somehow too complicated. Thanks.

jonathanmmm:
One of these pull requests has to be closed, either null or !value:
https://github.com/Mailtrain-org/mailtrain/pull/1154

PR #1057, opened 2021-05-04
Invalid dateformat default

MarcelWaldvogel:
The default refers to the non-existent `EUR` date format (should be `EU`; the string is `eur`, however).


bures:
Thanks.

PR #1058, opened 2021-05-04
Actually do ignore node_modules everywhere

MarcelWaldvogel:
`.dockerignore` matching rules differ from `.gitignore`, unfortunately. So, all `node_modules` directories (~750 MB) are sent to the docker daemon on every `docker build`.

Drive-by: Typo


bures:
Thanks.

Issue #1055, opened 2021-05-02
Duplicate translation keys


MarcelWaldvogel:
The translation files contain the following duplicate keys:
*   "disableClickedTracking"
*   "disableOpenedTracking"
*   "save"
*   "saveAndLeave"
*   "saving"
*   "sendConfiguration"
*   "sendConfiguration_plural"

I can gladly create a pull request to remove those (presuming that the entries further down in the file are later additions). However, I fear that they might reappear later.

What is the mechanism that has created those duplicate entries and is this mechanism still active (i.e., they will reappear)?

forgeOneHQ:
We are going to start with the development and testing of mailtrain v3 in the next weeks.

You are welcome to help us with the testing as soon as the first release candidate is available.

PR #1059, opened 2021-05-04
Translations dedup

MarcelWaldvogel:
Removes the duplicated keys from the translations files. Does not address the cause why the were added (see #1055).


bures:
It seems I have already merged something that conflicts with this PR. Could you please check? Thanks a lot.

MarcelWaldvogel:
Yes, indeed, you were incorporating PRs (thanks!!) while I tried to create a version which did not require the inclusion of my German translation update PR.

Redid it; however, as I do not know where these duplicate keys originated from, they may reappear later (I did not look into `extract.js` trying to find the cause).

bures:
Perfect. Thanks.

Issue #977, opened 2020-10-26
(v2) mailtrain won't connect to mysql when running via docker-compose


avataylor94:
when bringing up the service via docker-compose, 

i see the following on mysql's logs

```
mysql        | 2020-10-26  8:27:05 8 [Warning] Aborted connection 8 to db: 'unconnected' user: 'unauthenticated' host: '172.18.0.3' (This connection closed normally without authentication
```

mailtrain logs the following

```
mailtrain    | ERR! Queue/45 Could not initialize database: connect ECONNREFUSED 172.18.0.2:27017
mailtrain    | ERR! Child/default/45 [3f83045e93a7d158cafa] Sender process for default exited with 1
```

From a shell in the mailtrain instance

```
docker exec -it mailtrain bash
```

installed a mysql client
```
apk add mysql mariadb-client
```

connected via the client

```
mysql -h mysql -u mailtrain -pmailtrain 
```

seemed to work just fine

taenzerme:
@avataylor94 Have you found a solution for this? Same problem here.

saschafoerster:
I followed this issue before. It would make backups so much easier, if I could use the locally installed mariaDB, which has already some backup-scripts.

taenzerme:
I found a solution and will post the fixed configuration for the docker-compose.yml and the Mailtrain config file. Basically, the hostnames used in the default config (default.yaml: localhost, 127.0.0.1) don't work. Linking the containers in docker-compose to the mailtrain container and using the container names in the config file works.

minosimo:
I was previously able to run mailtrain with just `docker compose up`, but after stopping and restarting with no configuration changes I now also encounter this issue. I have the URL_BASE_TRUSTED, URL_BASE_SANDBOX and URL_BASE_PUBLIC variables set. Adding "depends_on" from taenzerme's pull request did not change the behavior.

forgeOneHQ:
We are going to start with the development and testing of mailtrain v3 in the next weeks.

You are welcome to help us with the testing as soon as the first release candidate is available.

PR #1093, opened 2021-06-11
Fixed issues with Zone-MTA to resolve dns

ghost:
Fixed issues with Zone-MTA regarding dns resolution by minimizing dns resolution time.
This will fix connection timed out errors and minimize mailtrain load time

```
mailtrain    | ERR! Queue/47 Could not initialize database: connection timed out
mailtrain    | ERR! Child/default/47 [] Sender process for default exited with 1
```

depends_on ensures all other container services are up before mailtrain container starts 
links ensure proper network connectivity for mailtrain container with rest of the container services

### DIFF
--- a/zone-mta/plugins/mailtrain-receiver.js
+++ b/zone-mta/plugins/mailtrain-receiver.js
@@ -13,7 +13,7 @@ module.exports.init = (app, done) => {
             envelope.dkim.keys = [];
         }
 
-        const dkimHeaderValue = headers.getFirst('x-mailtrain-dkim');
+        const dkimHeaderValue = require('libmime').decodeWords(headers.getFirst('x-mailtrain-dkim'));
 
         if (dkimHeaderValue) {
             const dkimKey = JSON.parse(dkimHeaderValue);


aamirmhussain:
This change will also solve the following issue
https://github.com/Mailtrain-org/mailtrain/issues/977

bures:
Thank you for the patch. I have doubts about why it is needed. The links section are just aliases and as per the documentation, they are not needed for services to communicate (https://docs.docker.com/compose/compose-file/compose-file-v2/#links). The depends_on only tells in which order services are started. But it does not wait for a service to be up and ready. Waiting for other services to become ready happens in https://github.com/Mailtrain-org/mailtrain/blob/v2/docker-entrypoint.sh . So from this perspective, I don't really understand the patch.

almereyda:
Yes, `links` (to other containers) are not needed, when two containers belong to the same docker network. Compose will also provide hostnames that map to a services definition, so they can be easily reached. This is solved via an internal DNS to Docker. Links were an early Docker feature, that have been suitable replaced by `depends_on` and `healthcheck`.

The https://docs.docker.com/compose/compose-file/compose-file-v2/#depends_on specification for a v2 Compose file knows a `condition: service_healthy` that will wait for a dependant service to pass its `healtcheck`, before attempting to run the container.

See https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck

Since the Compose v3 format used for Mailtrain v2 does not know the `condition: service_healthy` format, it comes to my mind to ask, if this PR could be reworked to introduce a dedicated user network for connecting the database services to the frontend, which is currently not present, and to follow the official guideline for maintaining a startup order, instead?

* [Networking in Compose | Docker Documentation](https://docs.docker.com/compose/networking/)
* [Control startup and shutdown order in Compose | Docker Documentation](https://docs.docker.com/compose/startup-order/)

PR #1103, opened 2021-06-23
Updated common.js for de-DE

usrflo:
Added missing translations for esp. servers/views/subscription/*


usrflo:
This pull request is replaced by https://github.com/Mailtrain-org/mailtrain/pull/1107

PR #1108, opened 2021-06-24
Bugfix of wrong/uninitialized var name (cachedFiles)

usrflo:
I think `fileCaches.delete(key);` is correct at this place to remove a cache file entry from the according cache map (while cachedFiles is not defined).

### DIFF
--- a/server/lib/file-cache.js
+++ b/server/lib/file-cache.js
@@ -152,7 +152,7 @@ async function _fileCache(typeId, cacheConfig, keyGen) {
                     if (fileStream) {
                         fileStream.destroy(err);
                         fs.unlink(tmpFilePath, () => {
-                            cachedFiles.delete(key);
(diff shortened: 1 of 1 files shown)


bures:
It has been long time since I wrote this cache. I believe the cachedFiles is a fragment forgotten there from the time when cache entries were not persisted in the DB. I believe that lines 154-156 should say something like:
```
                        fs.unlink(tmpFilePath, () => {
                            knex('file_cache').where('type', typeId).where('key', key).del().then(()=> callback());
                        });
```

Could you give it a try and if it works submit a PR? Thanks.

usrflo:
Your proposal didn't produce an error; but I cannot confirm if the cache deletion itself took place with your syntax.

Anyway, I replaced this PR with https://github.com/Mailtrain-org/mailtrain/pull/1109

PR #1109, opened 2021-06-24
Changed deletion of file from cache

usrflo:
Changed according to proposal in https://github.com/Mailtrain-org/mailtrain/pull/1108


bures:
Thanks. Let's give it a try. :-)

PR #1119, opened 2021-07-21
fix(language): add french in enabled language

PhilippeAccorsi:
Add french in dropdown button


PhilippeAccorsi:
Without this, it is not possible to easly activate french when working Mailtrain with docker.

bures:
Thanks.

PR #1154, opened 2021-09-29
Update tools.js

ganeshkrishnan1:
value is sometimes null and I get this error in logs

ERR! Senders Sending message to 3:4691 failed with error: Cannot read property 'replace' of null. Dropping the message.

### DIFF
--- a/server/lib/tools.js
+++ b/server/lib/tools.js
@@ -166,7 +166,7 @@ function _formatTemplateSimple(source, mergeTags, isHTML) {
             }
         }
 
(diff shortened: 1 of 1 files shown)


bures:
Thanks. In which situation does it happen that the value is null? Can you give an examples?

jonathanmmm:
One of these pull requests has to be closed, either null or !value:
https://github.com/Mailtrain-org/mailtrain/pull/1053/files

bures:
@jonathanmmm - good catch. I closed the other one. I believe we can resolve it here. I'll be happy to merge it, I just wanted to know the situation when it happens before we mask it.

PR #1175, opened 2022-01-01
Be able to set Default language in docker

jonathanmmm:
Only thing that is missing is enabledLanguages, but this should be the first step.
This way emails, like subscription are received in the correct language (don't know how to otherwise change the language of a subscription list)


jonathanmmm:
@bures a review would be great, are you the correct one for that?

jonathanmmm:
Closing in favour of: https://github.com/Mailtrain-org/mailtrain/pull/1177 as this is completely full of not needed commits

PR #1176, opened 2022-01-01
Security Risk: Warning about security risk and Update readme

jonathanmmm:
Warning about not setting ADMIN_PASSWORD being set to `test` if not set, being a great security hole.
And including ADMIN_PASSWORD and ADMIN_TOKEN into the table.
Don't know if ADMIN_ACCESS_TOKEN being empty makes it that none is set and optional or being another security hole.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,8 +238,12 @@ Alternatively, you can just declare them there leaving their value empty
 value can be provided via a file called `.env` or via environment 
 variables (e.g. `URL_BASE_TRUSTED=https://mailtrain.domain.com (and more env-vars..) docker-compose -f docker-compose.yml build (or up)`)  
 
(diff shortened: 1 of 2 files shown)


jonathanmmm:
@bures Please review, as this ENV was only found in docker-entrypoint.sh, why there admin password is changed and they don't know.
Big security problem. Maybe it should not be the default test, but more a random generated string, if not set or just not changed after restart and the one from databse used, only set to test, if no Admin user exists. But this is another issue. Right now users should be informed about the situation right now.

bures:
Makes sense. This is inspired by the mysql docker container. It is also supposed to be launched as `docker run --name some-mysql -e MYSQL_ROOT_PASSWORD=my-secret-pw -d mysql:tag`

I guess the problem is the default value. It should be a mandatory (similar to mysql).

PR #1177, opened 2022-01-01
Able to set default language via Docker ENV

jonathanmmm:
Only thing that is missing is enabledLanguages, but this should be the first step.
This way emails, like subscription are received in the correct language (don't know how to otherwise change the language of a subscription list)

@bures Who should be pinged for review?


jonathanmmm:
@bures ready for merging

bures:
Thanks.

PR #1180, opened 2022-01-02
make all TIMEZONEs lowercase via API

### LINKED ISSUE #1174: Bug TIMEZONE has to be written small via API
If you write e.g. Europe/Paris is not accepted and not error comes back. It will just set it in the UI to "Not Selected" (don't know what is under the hood).
As the examples show `eg. "Europe/Tallinn", "PST" or "UTC"` (actually it should be e.g., I think, typo).
I thought I will write the first letters capitalized. But this is wrong, you have to write all letters small, otherwise it won't be accepted, which is strange.

I propose to make just in the beginning of the API that the string of TIMEZONE is just made all small letters, so nobody has to worry about it. Or the examples are changed and is written, that it has to be all small letters.

jonathanmmm:
fixes that e.g. Europe/Paris is not accepted/shown in the UI, was it has to be written in only lower case letters.

fixed https://github.com/Mailtrain-org/mailtrain/issues/1174


bures:
This whole API is adopted from v1 and over the time become a bit too organic, so in the future, it would deserve some refactoring and better design. But for now, I believe, we can have it like this. Thanks for the patch.

PR #1238, opened 2022-04-17
Update api.js

l4b4r4b4b4:
Exclude `MERGE_FIRST_NAME` und `MERGE_LAST_NAME` from being lowercased to preserve user input.

### DIFF
--- a/server/routes/api.js
+++ b/server/routes/api.js
@@ -37,7 +37,11 @@ router.postAsync('/subscribe/:listCid', passport.loggedIn, async (req, res) => {
 
     const input = {};
     Object.keys(req.body).forEach(key => {
-        input[(key || '').toString().trim().toUpperCase()] = (req.body[key] || '').toString().trim().toLowerCase();
+        if (key=== "MERGE_FIRST_NAME" || key=== "MERGE_LAST_NAME" ) {
+            input[(key || '').toString().trim().toUpperCase()] = (req.body[key] || '').toString().trim()    
+        } else {
+            input[(key || '').toString().trim().toUpperCase()] = (req.body[key] || '').toString().trim().toLowerCase();
+        }
     });
 
     if (!input.EMAIL) {


bures:
The API is something I took over from Mailtrain v1. Overall, it's not in a good shape. I'm not sure why the lowercase is there, but we cannot remove it as other might depend on it. At the same time, the names of custom fields could be pretty much anything and you cannot rely on names to be MERGE_FIRST_NAME, etc. I'd suggest to add a parameter (e.g. "CASE_SENSITIVE"), which if you set to true, it won't do the lowercase.

l4b4r4b4b4:
Well I tried wether the other parameters would be affected. But couln't spot a difference. I don't quite get what you are trying to say depending custom fields concerning the matter^^ But yeah one could easy just imlement a CASE_SENSITIVE parameter to handle the input case.

bures:
Great. I just wanted to say that name doesn't have to be called MERGE_FIRST_NAME. So we should not rely on a set of predefined merge tags in the api.

l4b4r4b4b4:
Oh so what you want to say is, that we should change the name of the parameters to `FIRST_NAME` and `LAST_NAME` in the application in general, instead of adjusting everything to the given parameter names `MERGE_FIRST_NAME` and `MERGE_LAST_NAME`?

Well if I understood you correctly, I admit, this would be better. Will try to do it in the next days...

l4b4r4b4b4:
Since you said, the code is legacy from v1 and not in best shape, is there any general wish to refactor the api code? Would be willing to open a branch ;)

bures:
Indeed I had such a wish for a long time. But since I don't need the API for anything I do, it always had zero priority on my side. If you want to give it a try, I would be happy to help with consultation. I think it would make sense to first create some spec (e.g. on Apiary) and discuss it. I would like to consider it as brand new API. We may use API version identifier in request headers. That would allow us to come up with new behavior and still keep some level of backward compatibility.

l4b4r4b4b4:
Sure, if you're willing to help with consultation, I'd love to contribute!
Your suggestions sound like a legit approach. Can you recommend good resources to get started with apiary? I have not used it so far. 
I would start with building a API Spec inspired by the current api to have something concrete to discuss about.

bures:
Thanks. Great. I didn't use it either. It was just a suggestion based what I saw in other projects.

l4b4r4b4b4:
Well seems like a good tool to design, develop and test APIs. Will start in simply creating one for the status quo of mailtrain v2 external API in the next couple of days.

Do you use triggered campaigns a lot or at all? And does it work? I did not manage to fire up the triggers the other day.

l4b4r4b4b4:
Finally got to start working on the API doc in apiary.
Setup a project and synced it with my mailtrain fork on github.
https://lucashankedecansino.docs.apiary.io/#

PR #1240, opened 2022-04-17
Update subscription.js

l4b4r4b4b4:
Fixed bug, that caused redirect to wrong homepage url on subscription confirmation.


bures:
Why is this? Thanks.

PR #1245, opened 2022-04-25
Added extra SMTP sender throttling. 

tstrg:
The feature adjusts throttling options by allowing to set:

- The sender warm-up period gradually increases the volume each day until reaching your throttling settings.
- Enable/disable sender on specific weekdays

Currently implemented to SMTP sender only. Maybe someone will find it useful.

image

### DIFF
--- a/client/src/send-configurations/helpers.js
+++ b/client/src/send-configurations/helpers.js
@@ -51,9 +51,20 @@ export function getMailerTypes(t) {
         return {
             maxConnections: '5',
             throttling: '',
-            logTransactions: false
+            logTransactions: false,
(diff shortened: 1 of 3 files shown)


bures:
Good idea. Thanks. I guess you tested it, right?

tstrg:
@bures Yes, I've tested it locally, and seems all is working as expected. 

If someone wishes to test it with docker, the modification is available at https://hub.docker.com/repository/docker/tstrg/mailtrain

Btw I send greetings from the Czech rep.

bures:
Perfect. Thanks. Greetings from CZ too. :wink:

PR #1252, opened 2022-05-06
Fix issue #598

D4rk4:
@talheim-it this one fix issue with non-Latin chars

### DIFF
--- a/server/lib/tools.js
+++ b/server/lib/tools.js
@@ -10,7 +10,6 @@ const mjml2html = require('mjml');
 
 const hbs = require('hbs');
 const juice = require('juice');
-const he = require('he');
 const htmlToText = require('html-to-text');
 
 const fs = require('fs-extra');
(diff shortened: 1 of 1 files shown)


forgeOneHQ:
@D4rk4 thanks I forward it to @bures

bures:
I analyze the use of HE. I believe it is there to allow you to include text custom fields that contain characters that would be otherwise interpreted as HTML. For instance "<". So, to sum up, I think just removing the HE is not easily an option as then it means that you subscription fields have to be HTML-safe.

Just to be sure, I'm right, could you try in your modified Mailtrain instance the following FIRST_NAME: ```A <B> C``` ? Thanks.

D4rk4:
image
@bures

bures:
And what happens when you deliver an email to this person and use the FIRST_NAME merge tag in the campaign content?

D4rk4:
image
@bures in the message it's parsed as regular HTML tags. In the MT admin panel it's parsed as text.

bures:
Yes. This is what I thought. With the "simple" tag language, the tags should be interpreted as plain text. That's why the entity encoding is present. I believe, it would be better to to change the settings of HE (or use some other encoder), so that for instance <, >, & gets encoded, while the characters with various diacritics are not. Could you please give it a try? I would be happy to accept such PR. Thanks.

PR #1288, opened 2022-08-29
typo in server/routes/subscription.js

j-mir-prazak:
changed form.formInputStyle to form.form_input_style to match mysql database column - css styles should apply


bures:
Thanks.

Issue #1142, opened 2021-08-29
Mailtrain v2 - Update of server and client produce issue on zone-mta


forgeOneHQ:
I tried to update the mailtrain server and client and after all is running fine again, the mailing is not working anymore. Zone-MTA was not updated before, same issue after the updated, so I expect an issue  in between.

```
Aug 29 19:25:56 news01.example.com node[343885]: info HTTP PUT /rest/restricted-access-token 200 14.797 ms - -
Aug 29 19:25:57 news01.example.com node[346527]: undefined:1
Aug 29 19:25:57 news01.example.com node[346527]: =?UTF-8?Q?=7B=22domainName=22=3A=22example?= =?UTF-8?Q?=2Eat=22=2C=22keySelector=22=3A=222019>
Aug 29 19:25:57 news01.example.com node[346527]: SyntaxError: Unexpected token = in JSON at position 0
Aug 29 19:25:57 news01.example.com node[346527]:     at JSON.parse (<anonymous>)
Aug 29 19:25:57 news01.example.com node[346527]:     at Object.app.addHook [as action] (/opt/mailtrain/zone-mta/plugins/mailtrain-receiver.js:19:34)
Aug 29 19:25:57 news01.example.com node[346527]:     at Immediate.checkNext (/opt/mailtrain/zone-mta/node_modules/zone-mta/lib/plugin-handler.js:427:2>
Aug 29 19:25:57 news01.example.com node[346527]:     at runCallback (timers.js:705:18)
Aug 29 19:25:57 news01.example.com node[346527]:     at tryOnImmediate (timers.js:676:5)
Aug 29 19:25:57 news01.example.com node[346527]:     at processImmediate (timers.js:658:5)
Aug 29 19:25:57 news01.example.com node[343909]: ERR! SMTP/feeder/346527 Reciver process 9a4d24ffefb35f696515 for feeder exited with 1
Aug 29 19:25:57 news01.example.com node[344005]: ERR! Senders Sending message to 13:122 failed with error: write EPIPE. Dropping the message.
Aug 29 19:26:26 news01.example.com node[343885]: info HTTP PUT /rest/restricted-access-token 200 11.134 ms - -
Aug 29 19:26:34 news01.example.com node[346622]: undefined:1
[...]

PR #1292, opened 2022-09-03
decode possible encoded chunks in header

j-mir-prazak:
dkimKey expects JSON, however serveral times now, mailtrain started giving dkim header as encoded chunks. there is no try and catch on the json parse, so the emails won't be sent. 

require('libmime').decodeWords( header ) decodes chunks if present, or leaves the string alone - so even if the json is passed, it should not change the string.

i could not find the reason for the encoding so far

### DIFF
--- a/zone-mta/plugins/mailtrain-receiver.js
+++ b/zone-mta/plugins/mailtrain-receiver.js
@@ -13,7 +13,7 @@ module.exports.init = (app, done) => {
             envelope.dkim.keys = [];
         }
 
-        const dkimHeaderValue = headers.getFirst('x-mailtrain-dkim');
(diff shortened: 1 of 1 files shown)


bures:
Thanks. I've never encountered this problem. At least I'm not aware of it. Could we have one more person to test this PR on their installation before I merge it? Just to be sure this does not break anything.

iafilin (approved):

iafilin:
> Thanks. I've never encountered this problem. At least I'm not aware of it. Could we have one more person to test this PR on their installation before I merge it? Just to be sure this does not break anything.

@bures This change solves the problem #1142

bures:
Perfect. Thank you both.

PR #1301, opened 2022-09-28
Show link from options as a button in forms

saschafoerster:
via Slack and https://paste.sr.ht/blob/a2650eecd68e6da4e33e218314901c88143711d3


bures:
Thank you all.

